PR #109, opened 2026-04-08
feat(#108): add fallback-to-cache proxy mode with L1/L2 caching

### LINKED ISSUE #93: JSON fixture authoring guide
## Problem statement

Developers who want to use httptape for static mocking (without recording from a live API) have no documentation on how to hand-write Tape JSON fixture files. The Tape format is defined in code but not documented for manual authoring, making it hard to create fixtures from scratch or understand the structure of recorded files.

## User story

As a Go developer, I want documentation on the Tape JSON format so I can hand-write fixture files for static mocking without needing to record from a live API.

## Acceptance criteria

- [ ] `docs/json-fixture-authoring.md` guide covering:
  - Tape JSON structure with annotated field descriptions
  - Complete example fixture file for a typical REST API endpoint
  - Multiple examples: GET with JSON response, POST with request body, error responses, binary content
  - Tips for organizing fixtures in `testdata/` directories
  - How to use fixtures with `LoadFixtures()` and `LoadFixturesFS()`
- [ ] JSON Schema file for the Tape format (or reference the existing `config.schema.json` if it already covers this)
- [ ] Template fixture file in `docs/examples/` that developers can copy and modify
- [ ] All examples are valid JSON that can be loaded by httptape without modification

## Out of scope

- Auto-generation tooling for fixtures
- IDE plugins or editor integrations
- Fixture validation CLI command (future feature)

### LINKED ISSUE #97: UI-first dev guide
## Problem statement

Frontend developers who want to use httptape as a mock backend during development have no end-to-end documentation on the workflow. While individual features (CORS, latency, error simulation) will have their own docs, there is no guide that ties them together into a cohesive frontend development workflow.

## User story

As a frontend developer, I want a guide showing how to use httptape as a mock backend so I can develop my UI without a running backend.

## Acceptance criteria

- [ ] `docs/ui-first-dev-guide.md` covering:
  - Overview of the UI-first development workflow with httptape
  - Hand-writing fixture files for your API endpoints
  - Starting httptape as a mock server (CLI and Docker)
  - CORS setup for cross-origin frontend dev servers
  - Latency simulation for testing loading states
  - Error simulation for testing error handling
  - docker-compose example with a frontend dev server + httptape
  - Hot-reload workflow: editing fixtures and seeing changes immediately
  - Tips for organizing fixtures by feature/page
- [ ] Complete, runnable docker-compose example in `docs/examples/`
- [ ] All code examples are tested or copy-pasteable

## Out of scope

- Framework-specific guides (React, Vue, Angular, etc.)
- CI/CD integration for fixture validation
- Video tutorials or interactive demos

tibtof:
## Summary

- Add `docs/fixtures-authoring.md` (#93): comprehensive guide for hand-writing Tape JSON fixture files, covering the full JSON structure with field descriptions, working examples for GET/POST/204/paginated responses, metadata-based delay and error simulation, FileStore directory layout, and base64 encoding tips.
- Add `docs/ui-first-dev.md` (#97): practical guide for using httptape as a mock backend during frontend development, covering two workflows (hand-write vs record from staging), Docker Compose setup with a frontend service, CORS configuration (`--cors` flag), hot-reload behavior (edit fixture files without restarting), latency simulation for loading states, error simulation for error UI testing, and a comparison table with json-server and Mockoon.

Closes #93, closes #97

## Test plan

- [ ] Verify all JSON examples in fixtures-authoring.md are valid JSON
- [ ] Verify base64-encoded bodies decode to the documented JSON payloads
- [ ] Verify Docker Compose example in ui-first-dev.md is valid YAML
- [ ] Verify cross-references between the two new docs and existing docs resolve correctly
- [ ] Verify the docs render correctly on GitHub (markdown tables, code blocks, links)

### DIFF
--- a/cmd/httptape/main.go
+++ b/cmd/httptape/main.go
@@ -7,6 +7,7 @@
 //
 //	httptape serve    — Replay recorded fixtures as a mock HTTP server
 //	httptape record   — Proxy requests to upstream, record and sanitize responses
+//	httptape proxy    — Forward to upstream with L1/L2 fallback-to-cache
 //	httptape export   — Export fixtures to a tar.gz bundle
 //	httptape import   — Import fixtures from a tar.gz bundle
 package main
@@ -62,6 +63,7 @@ Usage:
 Commands:
   serve    Replay recorded fixtures as a mock HTTP server
   record   Proxy requests to upstream, record and sanitize responses
+  proxy    Forward to upstream with L1/L2 fallback-to-cache
   export   Export fixtures to a tar.gz bundle
   import   Import fixtures from a tar.gz bundle
 
@@ -91,6 +93,8 @@ func run(args []string) int {
 		err = runServe(cmdArgs)
 	case "record":
 		err = runRecord(cmdArgs)
+	case "proxy":
+		err = runProxy(cmdArgs)
 	case "export":
 		err = runExport(cmdArgs)
 	case "import":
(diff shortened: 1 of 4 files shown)


tibtof (commented):
## Architecture Review -- PR #109 (Proxy mode with L1/L2 caching)

**Reviewer**: Software Architect
**ADR**: ADR-26
**Verdict**: Approve with two required fixes (non-blocking -- both are small)

### ADR-26 Compliance

The implementation is faithful to ADR-26 across every major decision:

- **Separate `Proxy` type** in `proxy.go` rather than extending `Recorder` -- correct per SRP rationale
- **L1/L2 architecture**: L1 = `MemoryStore` (raw/ephemeral), L2 = `FileStore` (sanitized/persistent) -- matches ADR exactly
- **Constructor signature** `NewProxy(l1, l2 Store, opts ...ProxyOption)` with nil-panic guards -- matches ADR
- **Option naming** prefixed with `Proxy` (`WithProxyTransport`, `WithProxySanitizer`, etc.) -- avoids collision with Recorder options as specified
- **RoundTrip flow**: capture body, forward, isFallback check, dual-write (raw L1, sanitized L2), return real response -- matches ADR pseudocode
- **Fallback order**: L1 first, then L2, then original error -- correct
- **Synchronous writes** for both L1 and L2 -- matches ADR rationale
- **`X-Httptape-Source` header**: `l1-cache` / `l2-cache`, absent on success -- matches ADR
- **No `Close()` method** on Proxy -- correct per ADR lifecycle section
- **`isFallback` function** (not bool flag) with conservative default (transport errors only) -- matches ADR

### Security Boundary: Raw Never Hits Disk

**PASS**. The security invariant is correctly maintained:
- L1 receives raw (unsanitized) tapes and is always `MemoryStore` -- in-memory only
- L2 receives `p.sanitizer.Sanitize(rawTape)` before `p.l2.Save()` -- sanitization is applied before disk write
- CLI hardcodes `l1 := httptape.NewMemoryStore()` with no flag to override -- L1 is never configurable to FileStore
- `TestProxy_SanitizationOnL2Only` explicitly verifies L1 has raw auth header while L2 has it redacted

### CLI Integration

**PASS**. The `runProxy` function follows the established `runRecord` pattern:
- Required flags: `--upstream`, `--fixtures`
- Optional: `--config` (sanitizer for L2 only), `--port` (default 8081), `--cors`, `--fallback-on-5xx`
- Uses `httputil.ReverseProxy` with `Transport: tapeProxy` -- same pattern as record mode
- Graceful shutdown with signal handling -- matches existing commands
- Usage text and command routing updated

### Test Coverage

**PASS**. All 13 test cases from the ADR test plan are covered:
- Success path, L1 fallback, L2 fallback, no-match error, L1-before-L2 priority, sanitizer L2-only, 5xx fallback, X-Httptape-Source header (table-driven), no-Close contract, concurrent safety (50 goroutines with race detector), request body preserved for matching, nil-store panics, onError callback, route propagation
- All tests pass with `-race`
- `go vet` clean

### Issues Found

#### 1. Response body leak on 5xx fallback (Bug)

**File**: `proxy.go`, line 164-169

When `isFallback` returns true because of a 5xx response (not a transport error), `resp` is non-nil and its `Body` is never closed. The code jumps straight to `p.fallback(req, transportErr)` which returns a synthesized response from cache, abandoning the upstream response body.

**Fix**: Close `resp.Body` before entering fallback when `resp != nil`:
```go
if p.isFallback(transportErr, resp) {
    if resp != nil && resp.Body != nil {
        resp.Body.Close()
    }
    if reqBody != nil {
        req.Body = io.NopCloser(bytes.NewReader(reqBody))
    }
    return p.fallback(req, transportErr)
}
```

#### 2. `(nil, nil)` return violates `RoundTripper` contract on 5xx no-match (Bug)

**File**: `proxy.go`, line 246

When `isFallback` triggers on a 5xx (err is nil) and neither L1 nor L2 has a match, `fallback` returns `(nil, originalErr)` where `originalErr` is `nil`. This gives the caller `(nil, nil)` which violates the `http.RoundTripper` contract ("callers should treat a nil *Response with a nil error as if they received a non-nil *Response" -- but in practice most callers will panic on nil resp).

**Fix**: In the 5xx-fallback-no-match case, return the original 5xx response instead of `(nil, nil)`. This requires passing the original `resp` into `fallback`:
```go
func (p *Proxy) fallback(req *http.Request, originalErr error, originalResp *http.Response) (*http.Response, error) {
    // ... existing L1/L2 lookup ...

    // No match -- return original error or original response.
    if originalResp != nil {
        return originalResp, nil
    }
    return nil, originalErr
}
```
And do NOT close `resp.Body` before fallback -- only close it if fallback succeeds (i.e., found a cache hit). This way, on no-match the caller gets the real 5xx response back, which is the most useful behavior.

### Summary

The architecture is sound and the implementation is a clean, well-tested realization of ADR-26. The two bugs above are edge-case issues in the 5xx fallback path (the default transport-error-only path is unaffected). Both are small fixes. Approving with the expectation these are addressed before merge.

tibtof (commented):
## Review: PR #109 — Proxy mode with L1/L2 caching

Solid implementation overall. The architecture matches ADR-26 closely, the test coverage is thorough, and the code follows existing patterns (body capture, functional options, constructor guards). Two bugs need fixing before merge.

### Blocking issues

**1. Response body leak on 5xx fallback path** (`proxy.go`, `RoundTrip` method)

When `isFallback` returns `true` for a non-error response (e.g., 5xx with `WithProxyFallbackOn`), the upstream `resp.Body` is never closed. The code jumps straight to `p.fallback(req, transportErr)` without draining/closing the response body. This leaks the underlying TCP connection.

Fix: close `resp.Body` before entering the fallback path when `resp` is non-nil:

```go
if p.isFallback(transportErr, resp) {
    if resp != nil && resp.Body != nil {
        resp.Body.Close()
    }
    if reqBody != nil {
        req.Body = io.NopCloser(bytes.NewReader(reqBody))
    }
    return p.fallback(req, transportErr)
}
```

**2. `nil, nil` return violates `RoundTripper` contract** (`proxy.go`, `fallback` method)

When `isFallback` triggers on a 5xx (where `transportErr` is nil) and no cache match exists, `fallback` returns `return nil, originalErr` where `originalErr` is nil. This gives the caller `(nil, nil)`, which violates the `http.RoundTripper` contract ("a nil resp with a non-nil err, or a non-nil resp with a nil err").

Fix: when entering fallback from a 5xx (non-error) path, either pass the original 5xx response back as the fallback result, or synthesize a proper error. The cleanest approach is to change the `fallback` signature/call to also receive the original `resp`, and return it when no cache match is found:

```go
// In RoundTrip:
if p.isFallback(transportErr, resp) {
    ...
    return p.fallback(req, transportErr, resp)
}

// In fallback:
func (p *Proxy) fallback(req *http.Request, originalErr error, originalResp *http.Response) (*http.Response, error) {
    // ... L1/L2 lookups ...

    // No match — return originals.
    if originalResp != nil {
        return originalResp, nil  // return the 5xx as-is
    }
    return nil, originalErr
}
```

This way, on 5xx-triggered fallback with no cache hit, the caller gets the original 5xx response (which is still useful) rather than a contract violation. And on transport-error fallback with no cache hit, the caller gets the original error.

### Non-blocking observations

**3. Sanitizer/L1 data isolation is safe but fragile** (`proxy.go` lines ~115-120)

The code saves `rawTape` to L1, then passes the same `rawTape` value to `p.sanitizer.Sanitize(rawTape)` for L2. This works because `Tape` is a value type and the `SanitizeFunc` contract says "must not mutate the input." However, the `Headers` (map) and `Body` (slice) fields share underlying memory between the raw and sanitized copies. A defensive `rawTape.Request.Headers.Clone()` before the sanitize call would make this rock-solid, but I acknowledge this matches the existing Recorder pattern so it's consistent.

**4. Test for 5xx no-cache-match scenario is missing** (`proxy_test.go`)

There's a `TestProxy_FallbackOn5xx` test for the happy case (5xx triggers fallback, cache hit found), but no test for the unhappy case (5xx triggers fallback, no cache match). This is exactly the path where bug #2 manifests. Please add a test case.

**5. Minor: `TestProxy_OnErrorCallback` doesn't actually test error reporting**

The test sets up a proxy with a working transport and working stores, then asserts zero errors were reported. This verifies the callback doesn't fire spuriously but never exercises the error path. Consider adding a test with a store that returns errors on `Save` to verify errors are actually reported via the callback.

### What looks good

- L1/L2 dual-write ordering is correct (L1 first, then sanitize + L2)
- Fallback order is correct (L1 before L2, per ADR-26)
- Sanitization is correctly applied only to L2 writes, never L1
- `X-Httptape-Source` header values are correct ("l1-cache", "l2-cache", absent on success)
- Constructor guards (panic on nil stores) follow project conventions
- Request body capture and restore matches Recorder pattern exactly
- Concurrent safety test exists and exercises mixed success/failure
- CLI wiring follows existing patterns (flag parsing, reverse proxy, graceful shutdown)
- No external dependencies added (stdlib only)
- `TestProxy_XHttptapeSourceHeader` table-driven test correctly verifies L1 preference over L2

**Verdict: Request changes** — fix the two blocking bugs, add the missing test case, then this is ready to merge.

PR #113, opened 2026-04-10
feat: add TLS/mTLS support for upstream connections

tibtof:
## Summary

- Add `BuildTLSConfig` helper in `tls.go` that constructs `*tls.Config` from PEM file paths and an insecure flag (stdlib only)
- Add `WithProxyTLSConfig` and `WithRecorderTLSConfig` functional options for the Go library API
- Add `--tls-cert`, `--tls-key`, `--tls-ca`, `--tls-insecure` CLI flags to `proxy` and `record` subcommands
- Add comprehensive unit tests (cert/key validation, CA loading, insecure mode, error cases) and integration tests using `httptest.NewTLSServer`
- Add `docs/tls.md` with CLI examples, Go library examples, Docker guidance, and troubleshooting

Implements ADR-27. Closes #112.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./... -race` passes (all new and existing tests)
- [ ] Verify `BuildTLSConfig` returns `nil, nil` for zero args
- [ ] Verify cert-without-key and key-without-cert return errors
- [ ] Verify integration tests connect to TLS server with custom CA
- [ ] Verify insecure mode skips verification against TLS test server
- [ ] Verify nil TLS config is a no-op for both Proxy and Recorder

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #115, opened 2026-04-10
fix(#114): strip Content-Length from Proxy L2 fallback responses

### LINKED ISSUE #114: bug: Content-Length mismatch in Proxy L2 fallback responses
## Problem

When the Proxy falls back to L2 (disk) cached responses, the stored Content-Length header reflects the original (pre-sanitization) body size. The actual L2 body is different size due to redaction. This causes HTTP clients to reject the response (curl: "transfer closed with N bytes remaining", browser fetch API: "Failed to fetch").

This is the same issue that was fixed in Server.ServeHTTP (Content-Length is stripped before writing) but the Proxy's fallback response path doesn't strip it.

## Steps to reproduce

1. Start proxy with sanitization config
2. Make a request (populates L1 and L2)
3. Stop upstream + restart proxy (clears L1, forces L2)
4. Make the same request — Content-Length mismatch error

## Expected behavior

L2 fallback responses should have correct Content-Length (or no Content-Length, letting the HTTP server set it from actual body).

## Fix

Strip Content-Length header from L2 fallback responses in the Proxy's fallback method, same as Server.ServeHTTP does.

tibtof:
Closes #114

## Summary
Strip Content-Length header in Proxy.tapeToResponse() before returning cached responses. The stored header reflects the original body size, which differs after sanitization.

## Test plan
- [x] `go test ./... -race` passes
- [x] Manual test: L2 fallback now returns correct Content-Length matching actual body

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -311,6 +311,9 @@ func (p *Proxy) tapeToResponse(tape Tape, source string) *http.Response {
 		header = tape.Response.Headers.Clone()
 	}
 	header.Set("X-Httptape-Source", source)
+	// Remove stale Content-Length — the body may differ from the original
+	// due to sanitization. Let the HTTP stack set it from actual body size.
+	header.Del("Content-Length")
 
 	body := tape.Response.Body
 	if body == nil {


tibtof (commented):
## Review: feat: add TLS/mTLS support for upstream connections (PR #113)

Overall this is a well-structured PR that follows ADR-27 closely. `BuildTLSConfig` is correct, tests are thorough (unit + integration with real TLS servers), CLI wiring is clean, and docs are excellent. However, there is one **critical bug** that must be fixed before merge.

### MUST FIX: `WithProxyTLSConfig` / `WithRecorderTLSConfig` mutate `http.DefaultTransport`

Both option functions check `if t, ok := p.transport.(*http.Transport); ok` and then set `t.TLSClientConfig = cfg` directly on the matched transport. Since `NewProxy` and `NewRecorder` both initialize `transport` to `http.DefaultTransport` (which *is* an `*http.Transport`), this type assertion succeeds and the code **mutates the global `http.DefaultTransport`**.

This directly contradicts ADR-27 which states:

> `WithProxyTLSConfig` and `WithRecorderTLSConfig` do NOT clone `http.DefaultTransport`. When the user provides a TLS config, **a fresh `&http.Transport{TLSClientConfig: cfg}` is created**.

The fix: when the current transport is `http.DefaultTransport`, create a new `*http.Transport` instead of mutating it. For example:

```go
func WithProxyTLSConfig(cfg *tls.Config) ProxyOption {
    return func(p *Proxy) {
        if cfg == nil {
            return
        }
        if t, ok := p.transport.(*http.Transport); ok && t != http.DefaultTransport.(*http.Transport) {
            t.TLSClientConfig = cfg
            return
        }
        p.transport = &http.Transport{TLSClientConfig: cfg}
    }
}
```

Same pattern for `WithRecorderTLSConfig`.

The existing test `TestWithProxyTLSConfig_NilConfig` checks that nil is a no-op, and `TestWithProxyTLSConfig_CustomTransport` checks that a user-provided transport is reused. But there is no test that verifies **the default transport is NOT mutated** when TLS config is applied without a prior `WithProxyTransport` call. Please add a test like:

```go
func TestWithProxyTLSConfig_DoesNotMutateDefaultTransport(t *testing.T) {
    tlsCfg := &tls.Config{InsecureSkipVerify: true}
    l1 := NewMemoryStore()
    l2 := NewMemoryStore()
    _ = NewProxy(l1, l2, WithProxyTLSConfig(tlsCfg))

    dt := http.DefaultTransport.(*http.Transport)
    if dt.TLSClientConfig != nil {
        t.Fatal("WithProxyTLSConfig mutated http.DefaultTransport")
    }
}
```

### Everything else looks good

- **`BuildTLSConfig`**: Correct cert/key pairing validation, proper error wrapping with `%w`, CA pool handling validates PEM parse success, nil/nil return for zero args. All stdlib.
- **CLI flags**: Properly wired for both `record` and `proxy`. Warning logged for `--tls-insecure`. `usageError` wrapping is consistent with existing flags. Not added to `serve` (correct per ADR).
- **Test coverage**: All ADR-specified test cases are present -- no-args, cert-without-key, key-without-cert, insecure-only, cert+key, CA-only, all-options, non-existent files, invalid PEM. Integration tests use real `httptest.NewTLSServer`. Nil-config no-op tests for both Proxy and Recorder.
- **Docs**: `docs/tls.md` is comprehensive with all four levels, CLI examples, Go library examples, Docker guidance, and troubleshooting.

Requesting changes for the DefaultTransport mutation bug only. Everything else is ready.

PR #125, opened 2026-04-16
feat: technical health endpoint surface for proxy mode (#121)

### LINKED ISSUE #121: feat: technical health endpoint surface for proxy mode (snapshot + SSE + active probe)
## Problem

The proxy currently signals which tier served a request via the `X-Httptape-Source` response header (`l1-cache`, `l2-cache`, or absent for live upstream). This is per-request and reactive: a UI built on top of the proxy can only learn about an upstream outage (or recovery) by issuing a request and inspecting the header. The `ts-frontend-first` demo needs a way to update its "live / L1 / L2" badge **the moment** upstream state changes, without the user having to click Refresh.

## User story

As a Go developer embedding httptape in proxy mode, I want a small technical health surface (snapshot endpoint plus a server-sent events stream) that reports the proxy's current "what got served last" state and emits an event on every transition, so that operators and downstream UIs can react to upstream outages and recoveries in real time without polling per-request headers.

## Scope

A small technical surface mounted under the path prefix `/__httptape/`, available **only in proxy mode**, gated behind explicit opt-in flags so the default proxy behavior is unchanged.

### Endpoints

1. **`GET /__httptape/health`** — JSON snapshot of the current state.
2. **`GET /__httptape/health/stream`** — `text/event-stream` (SSE). Emits the current state immediately on connect (so reconnecting clients reseed without a race), then emits one event per state transition. Multiple concurrent subscribers must be supported.
3. **Active upstream probe** — a periodic background check (e.g. `HEAD /` or `GET /` against the configured upstream) so state transitions can be detected without a real client request driving them. Configurable interval. Off by default; auto-enabled at a sensible default cadence when the health endpoint is enabled and the probe interval is left unset.

### CLI flags (proposed names — refine if better fits exist; final names are the architect's call as long as the behavior matches)

- `--health-endpoint` (bool, default `false`) — when set, mounts both `/__httptape/health` and `/__httptape/health/stream` on the proxy listener.
- `--upstream-probe-interval` (duration, default `0` = disabled). When `--health-endpoint` is set and this flag is unset, the probe is auto-enabled at a sensible default cadence (e.g. ~2s — pick a value the architect can defend in the design).

### Library API

The same capability must be reachable from the embeddable Go API (functional options on `NewProxy` or equivalent — the architect decides the surface), so users embedding the library directly get the same behavior without going through the CLI binary. The CLI flags are a thin wrapper on top.

## Definition of "state" (must be settled in the spec, with reasoning)

Two natural definitions exist:

- **A — upstream health**: `live` if the last probe succeeded, `degraded` otherwise. L1/L2 then becomes a separate per-request signal.
- **B — most-recently-served source**: reflects what the proxy actually served on its last handled request (real upstream, L1, or L2). The active probe is treated like any other request through the proxy's resolution path, so a probe success looks like a `live` serve and a probe failure followed by an L2 hit looks like an `l2-cache` serve.

**Settle on definition B.** Rationale: the consuming UI tracks what the user is *experiencing*. When the upstream is down and the proxy is serving from L2, "degraded" is the wrong story — the user is in fact getting a successful L2 response and should see the badge reflect that. Treating the probe as a synthetic request through the same resolution path means there is one source-of-truth state machine, not two parallel ones. State values are aligned with the existing `X-Httptape-Source` semantics:

- `live` — last serve came from the real upstream (header was absent).
- `l1-cache` — last serve came from L1 fallback.
- `l2-cache` — last serve came from L2 fallback.

If neither cache matched and the original transport error was returned, the state remains whatever it was before that failed serve. The architect may refine the exact transition rules but should keep the "single source of truth, fed by both real and probe traffic" property.

## Acceptance criteria

Behavior-based; each must be testable without coordination.

### Snapshot endpoint
- [ ] When `--health-endpoint` is **not** set, `GET /__httptape/health` returns 404 (or whatever the default proxy handler returns) — the endpoint must not be mounted.
- [ ] When `--health-endpoint` **is** set, `GET /__httptape/health` returns HTTP 200 with `Content-Type: application/json` and a body containing at least: current `state` (one of `live`, `l1-cache`, `l2-cache`), the configured `upstream_url`, the timestamp of the most recent probe (`last_probed_at`, RFC 3339; null/absent when probing is disabled or no probe has run yet), and the `probe_interval_ms` in effect (0 when disabled).
- [ ] The snapshot reflects the same state value the SSE stream would emit at the same instant (no skew between the two surfaces).
- [ ] The snapshot endpoint is safe to call concurrently from many clients.

### SSE stream endpoint
- [ ] `GET /__httptape/health/stream` returns HTTP 200 with `Content-Type: text/event-stream` and disables proxy buffering (`Cache-Control: no-cache`, appropriate connection handling).
- [ ] On connect, the server immediately emits one event carrying the current state (so a reconnecting client does not have to wait for the next transition to learn where it stands).
- [ ] After the initial event, exactly one event is emitted per state transition. No event is emitted when a serve confirms the existing state (e.g. two consecutive `live` serves emit nothing after the initial event).
- [ ] At least 100 concurrent SSE subscribers can be served from a single proxy without dropped events for any subscriber under steady-state operation.
- [ ] A slow or dead subscriber must not block other subscribers and must not block the proxy's request path or the probe loop. The architect chooses the back-pressure strategy (drop, disconnect, bounded buffer) and documents it.
- [ ] When the proxy shuts down, all open SSE connections are closed cleanly (the client sees EOF, not a hung socket).

### Active upstream probe
- [ ] When `--upstream-probe-interval` is `0` and `--health-endpoint` is not set, no probe goroutine runs.
- [ ] When `--health-endpoint` is set and `--upstream-probe-interval` is unset, a probe runs at a sensible default cadence (documented in the design).
- [ ] When `--upstream-probe-interval` is set explicitly, that interval is honored (within reasonable scheduling tolerance).
- [ ] The probe goes through the proxy's normal resolution path so its outcome feeds the same state machine that real requests do (per definition B above).
- [ ] The probe respects the configured TLS/mTLS settings (`--tls-cert`, `--tls-key`, `--tls-ca`, `--tls-insecure`) — if the existing proxy options apply, the probe must apply them too.
- [ ] On proxy shutdown, the probe loop exits within the existing graceful-shutdown window and does not leak goroutines.

### Concurrency and safety
- [ ] All shared state (current state, subscriber set, last-probe metadata) is safe for concurrent use by the request path, the probe loop, and the SSE handler. The full test suite must pass under `go test -race`.
- [ ] No package-level state, no `init()`, no side effects on import (per CLAUDE.md).
- [ ] No new external dependencies — stdlib only.
- [ ] No panics in the request path, the probe loop, or the SSE handler. Constructor-guard panics on nil required dependencies are acceptable per existing convention.

### Backward compatibility
- [ ] With both flags at their defaults, proxy behavior is byte-for-byte identical to today (no new endpoints mounted, no new goroutines, no new headers, `X-Httptape-Source` still emitted as before).
- [ ] The `X-Httptape-Source` response header is **not** removed or renamed by this change. It remains the per-request ground truth.

### Tests
- [ ] Unit tests live next to the code they test (`*_test.go`), table-driven where natural.
- [ ] Test suite stays race-clean (`go test -race ./...`).
- [ ] Coverage for the new code meets the project's 90% target.
- [ ] At least one integration-style test exercises: probe detects upstream-down, SSE subscriber receives the transition event, probe detects upstream-up, SSE subscriber receives the recovery event — all without issuing a real client request between transitions.
- [ ] At least one test verifies graceful shutdown closes SSE subscribers and stops the probe loop within the shutdown window.

### Documentation
- [ ] Both new flags are documented in the proxy command's `--help` output.
- [ ] The endpoint paths, response shapes, and SSE event format are documented in package godoc and (briefly) in the README's proxy section.

## Non-goals (explicitly out of scope)

- Authentication, authorization, or rate-limiting on the health endpoints. They are intended to live behind a trusted boundary; gating them is an operator concern.
- Request-level metrics (latency histograms, per-route counters, request counts). This issue is about state, not telemetry.
- Prometheus / OpenMetrics exposition. A separate future feature.
- Liveness/readiness semantics for Kubernetes — the snapshot is suitable for ad-hoc probes but this issue does not commit to specific HTTP-status conventions for k8s integration.
- Server-Sent Events `Last-Event-ID` / replay history. The initial-on-connect event is the resync mechanism; no event log is retained.
- WebSocket transport. SSE is sufficient and one-way.
- Surfacing the state in `serve`, `record`, or `mock` modes. Proxy mode only — the concept of "what tier served" is proxy-specific.
- Per-session or per-visitor outage simulation (the toggle endpoint sketched in `ts-frontend-first/BACKLOG.md`). That is a separate downstream feature that depends on this one landing first.
- Changing `X-Httptape-Source` values, the L1/L2 cache layout, the matcher, or the sanitizer.

## Locked constraints (from CLAUDE.md — must respect)

- **stdlib only.** SSE is implementable directly on `net/http` via `http.Flusher` and manual `data: ...\n\n` framing.
- **No package-level state, no `init()`, no side effects on import.** Everything via dependency injection / functional options.
- **Functional options pattern** for any new public constructor inputs.
- **Hexagonal-by-convention.** Fits within the existing single-package layout (e.g. a new `health.go` plus `health_test.go` alongside `proxy.go`). No `internal/` reorganization.
- **Graceful shutdown.** Probe loop and active SSE connections must drain on proxy close within the existing shutdown window.
- **No panics in hot paths.** Constructor guards on nil dependencies are the only acceptable panic.
- **Tests next to code, table-driven where natural, race-clean, 90% coverage target.**
- **Embeddable first.** The library API must be usable without going through the CLI.

## Downstream consumer (context only — not part of this issue)

Once this lands, `ts-frontend-first` will:
- Add an `EventSource("/__httptape/health/stream")` in the frontend.
- On every state event, refetch data and update the badge.
- Stop driving the badge from the per-request `X-Httptape-Source` header (the header stays as ground truth, just no longer drives UI).

A second feature (per-visitor outage simulation, tracked in `~/workspace/httptape-demos/ts-frontend-first/BACKLOG.md`) depends on this one landing first but is **out of scope** here.

- Final names for the two CLI flags (proposed: `--health-endpoint`, `--upstream-probe-interval`).
- Default probe cadence when `--health-endpoint` is enabled and the interval is unset (suggested ~2s — defend or revise in the design).
- Probe HTTP method and path against upstream (`HEAD /` vs `GET /` vs configurable). Smallest-blast-radius default preferred.
- Back-pressure strategy for slow/dead SSE subscribers (drop event, disconnect subscriber, bounded buffer per subscriber). Document trade-off in the design.
- Exact JSON field names and shape for the `/__httptape/health` response (the field list above is a suggestion; the architect picks final names and confirms shape).
- Whether the SSE event uses an `event:` name (e.g. `event: state`) or relies on the default `message` event. Pick one and document it.

tibtof:
Closes #121

## Summary

Adds an opt-in technical surface to proxy mode for real-time observation of the
"most-recently-served source" state machine:

- `GET /__httptape/health` — JSON snapshot.
- `GET /__httptape/health/stream` — `text/event-stream` (initial-on-connect + one event per state transition).
- Optional active probe (`HEAD /` with sticky `GET /` promotion on 405/501) that goes through `Proxy.RoundTrip` so it feeds the same state machine real client traffic does.

State values (`live`, `l1-cache`, `l2-cache`) mirror the existing `X-Httptape-Source` response header semantics. The header itself is preserved unchanged.

CLI flags: `--health-endpoint`, `--upstream-probe-interval` (defaults to 2s when the endpoint is enabled and the interval is unset). Library API: `WithProxyHealthEndpoint`, `WithProxyProbeInterval`, `WithProxyProbePath`, `WithProxyHealthErrorHandler`, `WithProxyUpstreamURL` plus `Proxy.HealthHandler`, `Proxy.Start`, `Proxy.Close`.

### Backward compatibility (locked)

With both flags absent / both options unset:

- `Proxy.HealthHandler()` returns `nil`.
- `Proxy.Start()` and `Proxy.Close()` are no-ops.
- The request path adds a single nil-receiver method call (`p.health.observe(...)`) which the compiler typically inlines away.
- No new headers, no new goroutines, no behavior change. Locked by `TestProxy_HealthDisabledByDefault` and the existing proxy tests.

### Concurrency model (per ADR-28)

- Single `sync.Mutex` guards `(state, since, lastProbed, subscribers)`.
- Per-subscriber bounded channel (size 8); on overflow the broadcaster drops and disconnects the subscriber, the SSE handler returns, and `EventSource` reconnects with a fresh initial seed.
- Probe loop derived from a `done` channel; `Close` cancels in-flight probe RoundTrips immediately and waits for the goroutine.
- All tests pass under `go test -race`.

## Sample output

`GET /__httptape/health`:
```
{"state":"live","upstream_url":"http://127.0.0.1:19999","last_probed_at":"2026-04-16T16:59:59.781417Z","probe_interval_ms":500,"since":"2026-04-16T16:59:58.773999Z"}
```

`GET /__httptape/health/stream`:
```
retry: 2000

data: {"state":"live","upstream_url":"http://127.0.0.1:19999","last_probed_at":"2026-04-16T16:59:59.781417Z","probe_interval_ms":500,"since":"2026-04-16T16:59:58.773999Z"}

```

## Test plan

- [x] `go test ./... -race` clean
- [x] `go vet ./...` clean
- [x] `gofmt -l` clean on touched files
- [x] Backward-compat regression (`TestProxy_HealthDisabledByDefault`): default-off proxy exposes no health routes, starts no goroutines, emits no new headers.
- [x] `X-Httptape-Source` still emitted on cache fallback when health endpoint is enabled (`TestProxy_HealthHeaderUnchanged`).
- [x] Integration: probe drives live -> l1-cache -> live transitions; SSE subscriber receives both events without any client request between transitions (`TestProxy_HealthIntegration`).
- [x] Slow subscriber dropped after buffer overflow (`TestHealthMonitor_SlowSubscriberDropped`).
- [x] 100 concurrent subscribers each receive transition events (`TestHealthMonitor_BroadcastFanOut`).
- [x] Graceful shutdown closes SSE streams (`TestHealthMonitor_HTTPStreamGracefulShutdown`).
- [x] Probe loop exits on Close, no goroutine leaks (`TestHealthMonitor_CloseStopsProbe`, `TestProxy_StartCloseIdempotent`).
- [x] HEAD->GET sticky promotion on 405 (`TestHealthMonitor_ProbeHEADtoGETPromotion`).
- [x] CLI flags (`TestProxyHelpExposesHealthFlags`, `TestProxyUpstreamProbeIntervalRequiresHealthEndpoint`, `TestProxyHealthEndpointMounted`).
- [x] Coverage on `health.go` averages well above 90% (most functions 100%).

## Files

- `health.go`, `health_test.go` (new)
- `proxy.go`, `proxy_test.go`, `cmd/httptape/main.go`, `cmd/httptape/main_test.go`, `doc.go`, `README.md`, `decisions.md` (modified — `decisions.md` carries ADR-28)

### DIFF
--- a/health.go
+++ b/health.go
@@ -215,8 +215,13 @@ func (h *HealthMonitor) recordProbeAttempt() {
 }
 
 // applyObservation runs the observe algorithm: detect transition, update
-// state under the lock, snapshot subscribers, then send out events without
-// holding the lock.
+// state under the lock, then broadcast to subscribers while still holding the
+// lock. Holding h.mu during the send is what protects against the
+// send-on-closed-channel race: dropSubscriber (the single owner of close) also
+// requires h.mu, so a subscriber's channel cannot be closed mid-send. The
+// per-subscriber buffer is bounded; select/default keeps the broadcast
+// non-blocking and overflowed subscribers are dropped in-place under the same
+// lock.
 func (h *HealthMonitor) applyObservation(src SourceState, fromProbe bool) {
 	if !validSourceState(src) {
 		return
@@ -225,6 +230,8 @@ func (h *HealthMonitor) applyObservation(src SourceState, fromProbe bool) {
 	now := h.now().UTC()
 
 	h.mu.Lock()
+	defer h.mu.Unlock()
+
 	transitioned := src != h.state
 	if transitioned {
 		h.state = src
@@ -236,24 +243,30 @@ func (h *HealthMonitor) applyObservation(src SourceState, fromProbe bool) {
 	}
 
 	if !transitioned {
-		h.mu.Unlock()
 		return
 	}
 
 	snap := h.snapshotLocked()
-	subs := make([]*healthSubscriber, 0, len(h.subscribers))
-	for sub := range h.subscribers {
-		subs = append(subs, sub)
-	}
-	h.mu.Unlock()
+	h.broadcastLocked(snap)
+}
 
-	for _, sub := range subs {
+// broadcastLocked sends snap to every subscriber non-blockingly. Subscribers
+// whose buffer is full are dropped (channel closed, removed from the set)
+// under the same lock. Caller MUST hold h.mu.
+func (h *HealthMonitor) broadcastLocked(snap HealthSnapshot) {
+	var toDrop []*healthSubscriber
+	for sub := range h.subscribers {
 		select {
 		case sub.ch <- snap:
 		default:
-			// Bounded buffer overflow: drop and disconnect.
-			h.dropSubscriber(sub)
+			// Bounded buffer overflow: schedule a drop after we finish iterating
+			// (don't mutate the map while ranging over it).
+			toDrop = append(toDrop, sub)
 		}
+	}
+	for _, sub := range toDrop {
+		delete(h.subscribers, sub)
+		close(sub.ch)
 	}
 }
 
@@ -282,21 +295,34 @@ func (h *HealthMonitor) snapshotLocked() HealthSnapshot {
 
 // subscribe registers a new SSE subscriber and returns its receive channel
 // plus an unsubscribe func. The channel is closed when the subscriber is
-// dropped (either by overflow or by Close).
+// dropped (either by overflow or by Close, which goes through dropSubscriber).
+//
+// The initial seed is delivered while holding h.mu. The channel is fresh with
+// capacity sseBufferSize so the send is guaranteed not to block; holding the
+// lock makes the registration + seed atomic with respect to broadcast and
+// dropSubscriber, eliminating any send-on-closed-channel window.
 func (h *HealthMonitor) subscribe() (<-chan HealthSnapshot, func()) {
 	sub := &healthSubscriber{ch: make(chan HealthSnapshot, sseBufferSize)}
 
 	h.mu.Lock()
+	// Reject post-Close subscribes deterministically: return a closed channel
+	// and a no-op unsub so callers don't block forever.
+	select {
+	case <-h.done:
+		h.mu.Unlock()
+		close(sub.ch)
+		return sub.ch, func() {}
+	default:
+	}
 	snap := h.snapshotLocked()
 	h.subscribers[sub] = struct{}{}
-	h.mu.Unlock()
-
-	// Initial seed: the channel is fresh and capacity is sseBufferSize, so
-	// this never blocks. We still use select/default for paranoia.
+	// Initial seed: send under the lock so dropSubscriber cannot race us. The
+	// buffer is fresh, so this cannot block; select/default is defensive.
 	select {
 	case sub.ch <- snap:
 	default:
 	}
+	h.mu.Unlock()
 
 	unsub := func() {
 		h.dropSubscriber(sub)
--- a/health.go
+++ b/health.go
@@ -344,11 +344,14 @@ func (h *HealthMonitor) runProbe() {
 	defer ticker.Stop()
 
 	// Probe context cancels when the monitor is closed so any in-flight
-	// RoundTrip is aborted promptly.
+	// RoundTrip is aborted promptly. The watcher goroutine is tracked in wg
+	// so Close() waits for it to exit before returning.
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	h.wg.Add(1)
 	go func() {
+		defer h.wg.Done()
 		select {
 		case <-h.done:
 			cancel()
@@ -361,7 +364,16 @@ func (h *HealthMonitor) runProbe() {
 		case <-h.done:
 			return
 		case <-ticker.C:
-			h.runProbeOnce(ctx)
+			// Re-check h.done non-blockingly: when both <-h.done and <-ticker.C
+			// are ready, Go's select picks at random. Without this guard a
+			// probe could fire after Close() has already been observed, which
+			// breaks the "no probes after Close" contract.
+			select {
+			case <-h.done:
+				return
+			default:
+				h.runProbeOnce(ctx)
+			}
 		}
 	}
 }
@@ -389,13 +401,15 @@ func (h *HealthMonitor) runProbeOnce(ctx context.Context) {
 
 	resp, err := h.transport.RoundTrip(req)
 	if err != nil {
-		// Transport error AND no response: state stays as-is per definition B.
-		// The Proxy itself may have called observe(...) on a cache hit before
-		// returning an error path; we don't second-guess that.
-		if resp == nil {
-			h.reportError(fmt.Errorf("httptape: probe transport: %w", err))
-			return
+		// http.RoundTripper allows (resp != nil, err != nil) — e.g. partial
+		// body reads. We always close the body if present to avoid leaking
+		// connections, surface the error, and return: the response cannot be
+		// trusted to drive the state machine in this case.
+		if resp != nil && resp.Body != nil {
+			resp.Body.Close()
 		}
+		h.reportError(fmt.Errorf("httptape: probe transport: %w", err))
+		return
 	}
 	if resp == nil {
 		return
--- a/health.go
+++ b/health.go
@@ -1,0 +1,580 @@
+package httptape
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+)
+
+// SourceState identifies which tier produced the most recent serve through
+// the proxy. Values mirror the X-Httptape-Source response header semantics
+// established in ADR-26.
+type SourceState string
+
+const (
+	// StateLive indicates that the last serve came from the real upstream.
+	// On the per-request response, this corresponds to the absence of an
+	// X-Httptape-Source header.
+	StateLive SourceState = "live"
+	// StateL1Cache indicates that the last serve came from the L1 (in-memory)
+	// fallback cache. Mirrors X-Httptape-Source: l1-cache.
+	StateL1Cache SourceState = "l1-cache"
+	// StateL2Cache indicates that the last serve came from the L2 (on-disk)
+	// fallback cache. Mirrors X-Httptape-Source: l2-cache.
+	StateL2Cache SourceState = "l2-cache"
+)
+
+// HealthSnapshot is the JSON payload returned by GET /__httptape/health and
+// emitted on every SSE event on /__httptape/health/stream. The snapshot
+// endpoint and the SSE stream agree byte-for-byte at any instant — both
+// surfaces read the same protected state.
+type HealthSnapshot struct {
+	// State is the current "most-recently-served source".
+	State SourceState `json:"state"`
+	// UpstreamURL is the configured upstream URL the proxy targets.
+	UpstreamURL string `json:"upstream_url"`
+	// LastProbedAt is the timestamp of the most recent active probe attempt
+	// (RFC 3339). Nil/omitted when probing is disabled or no probe has run yet.
+	LastProbedAt *time.Time `json:"last_probed_at,omitempty"`
+	// ProbeIntervalMS is the configured probe cadence in milliseconds.
+	// Zero when probing is disabled.
+	ProbeIntervalMS int64 `json:"probe_interval_ms"`
+	// Since is when the proxy last transitioned into the current state
+	// (RFC 3339).
+	Since time.Time `json:"since"`
+}
+
+// HealthMonitor owns the proxy's "most-recently-served source" state, the SSE
+// subscriber set, and the active probe loop. It is created by the Proxy when
+// WithProxyHealthEndpoint is enabled. With the option absent, the Proxy holds
+// a nil *HealthMonitor and the request path takes a fast no-op branch — this
+// preserves byte-for-byte default behavior.
+//
+// HealthMonitor is safe for concurrent use.
+type HealthMonitor struct {
+	upstreamURL string
+	interval    time.Duration // 0 = no probe loop
+	probePath   string
+	transport   http.RoundTripper
+	onError     func(error)
+	now         func() time.Time
+
+	probeMethodMu sync.Mutex
+	probeMethod   string // dynamically promoted from HEAD to GET on 405/501
+
+	// state machine + subscriber set
+	mu          sync.Mutex
+	state       SourceState
+	since       time.Time
+	lastProbed  *time.Time
+	subscribers map[*healthSubscriber]struct{}
+
+	// lifecycle
+	startOnce sync.Once
+	closeOnce sync.Once
+	done      chan struct{}
+	wg        sync.WaitGroup
+}
+
+// HealthMonitorOption configures a HealthMonitor.
+type HealthMonitorOption func(*HealthMonitor)
+
+// healthSubscriber is one connected SSE client. The buffer is bounded; if a
+// send would block (buffer full) the broadcast routine drops the subscriber,
+// closes the buffer, and the handler goroutine exits writing EOF to the
+// underlying connection.
+type healthSubscriber struct {
+	ch chan HealthSnapshot
+}
+
+const (
+	sseBufferSize        = 8 // per-subscriber bounded buffer
+	defaultProbeInterval = 2 * time.Second
+	defaultProbePath     = "/"
+	defaultProbeMethod   = http.MethodHead
+	healthEndpointPath   = "/__httptape/health"
+	healthStreamPath     = "/__httptape/health/stream"
+	sseRetryBackoffMS    = 2000
+	probeHeaderName      = "X-Httptape-Probe"
+)
+
+// WithHealthClock injects a clock for tests. Defaults to time.Now.
+func WithHealthClock(now func() time.Time) HealthMonitorOption {
+	return func(h *HealthMonitor) {
+		if now != nil {
+			h.now = now
+		}
+	}
+}
+
+// WithHealthInterval sets the active probe cadence on the HealthMonitor
+// directly. Library users typically configure this through
+// WithProxyProbeInterval; this option exists so callers constructing a
+// HealthMonitor by hand (tests, advanced embedding) can set the interval.
+func WithHealthInterval(d time.Duration) HealthMonitorOption {
+	return func(h *HealthMonitor) {
+		if d < 0 {
+			d = 0
+		}
+		h.interval = d
+	}
+}
+
+// WithHealthProbePath sets the URL path the active probe targets on the
+// upstream (default "/"). Library-only knob.
+func WithHealthProbePath(path string) HealthMonitorOption {
+	return func(h *HealthMonitor) {
+		if path != "" {
+			h.probePath = path
+		}
+	}
+}
+
+// WithHealthErrorHandler sets a callback for non-fatal errors inside the
+// health surface (probe transport errors that are not state transitions,
+// SSE write errors, etc.).
+func WithHealthErrorHandler(fn func(error)) HealthMonitorOption {
+	return func(h *HealthMonitor) {
+		if fn != nil {
+			h.onError = fn
+		}
+	}
+}
+
+// NewHealthMonitor constructs a HealthMonitor. transport is the
+// http.RoundTripper the active probe will hit — in production this is the
+// Proxy itself, so the probe takes the same resolution path as real client
+// traffic. upstreamURL is reported in the snapshot. opts may inject a clock,
+// configure the probe interval, or provide an error callback.
+//
+// upstreamURL must be non-empty and transport must be non-nil. Both are
+// programming errors and panic per the constructor-guard convention.
+func NewHealthMonitor(upstreamURL string, transport http.RoundTripper, opts ...HealthMonitorOption) *HealthMonitor {
+	if upstreamURL == "" {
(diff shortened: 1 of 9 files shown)


health.go
```diff
@@ -0,0 +1,580 @@
...
+	}()
+
+	for {
+		select {
+		case <-h.done:
+			return
+		case <-ticker.C:
+			h.runProbeOnce(ctx)
+		}
+	}
```
tibtof, line 404: **Blocking — flaky concurrency contract.** `TestHealthMonitor_CloseStopsProbe` is flaky: failing ~14/20 runs under `-race -count=20` and ~4/20 runs without race. Root cause is in `runProbe`: when both `<-h.done` and `<-ticker.C` are ready, `select` picks one at random, so a probe can fire after `Close()` has already been called. The test asserts "no further probes after Close" — exactly the property the implementation cannot guarantee today.

Two reasonable fixes:
1. Re-check `<-h.done` before invoking `runProbeOnce`:
```go
case <-ticker.C:
    select {
    case <-h.done:
        return
    default:
        h.runProbeOnce(ctx)
    }
```
2. Pass the cancelled-on-Close `ctx` into the transport and have callers honor it; let the recordingTransport in tests simulate that.

Option 1 is the smaller surgery. Either way, the contract "no probes after Close" needs to hold deterministically — otherwise this is a real surprise for embedders relying on `Close()`.

health.go
```diff
@@ -0,0 +1,580 @@
...
+	snap := h.snapshotLocked()
+	h.subscribers[sub] = struct{}{}
+	h.mu.Unlock()
+
+	// Initial seed: the channel is fresh and capacity is sseBufferSize, so
+	// this never blocks. We still use select/default for paranoia.
+	select {
+	case sub.ch <- snap:
+	default:
+	}
```
tibtof, line 324: **Blocking — send-on-closed-channel race.** `subscribe()` registers the subscriber under `h.mu`, releases the lock, and then sends the initial seed via `select { case sub.ch <- snap: default: }`. The `select/default` pattern only guards against a *full* channel; it does NOT guard against a *closed* channel. Sending on a closed channel panics — even inside select.

Sequence that crashes:
1. T1 `subscribe`: holds h.mu, registers `sub`, releases h.mu.
2. T2 `Close`: acquires h.mu, iterates `subscribers` (sees `sub`), `close(sub.ch)`, releases h.mu.
3. T1: enters the select/default; `sub.ch <- snap` panics.

The same hazard exists in `applyObservation` (the snapshot-then-broadcast pattern is identical).

Suggested fix: either (a) check `h.done` *and* hold `h.mu` while sending, or (b) gate the send on a per-monitor "closed" flag, or (c) have `Close` not close the subscriber channels and instead let the SSE handlers exit on `<-h.done` and rely on `unsub` to close the channel exactly once. A single-owner-of-close pattern is the cleanest.

Today the existing tests don't trigger this because no test races `subscribe` against `Close`. Add a stress test (e.g. N goroutines doing subscribe/unsub while another calls Close) — that will catch it deterministically under `-race`.

health.go
```diff
@@ -0,0 +1,580 @@
...
+	h.mu.Unlock()
+
+	for _, sub := range subs {
+		select {
+		case sub.ch <- snap:
+		default:
+			// Bounded buffer overflow: drop and disconnect.
+			h.dropSubscriber(sub)
+		}
+	}
```
tibtof, line 266: Same close/broadcast race risk as the one I flagged in `subscribe`: this loop captures `subs` under `h.mu` and then sends without the lock. If `Close` runs between the `Unlock` and these sends, the channels in `subs` are already closed and `sub.ch <- snap` panics. The current tests don't exercise the timing window, but it's reachable.

Pairing the broadcast with a single-owner-of-close convention (only `unsub` closes the channel; `Close` signals via `h.done` and waits for handlers) eliminates both this and the subscribe-side race in one move.

health.go
```diff
@@ -0,0 +1,580 @@
...
+	resp, err := h.transport.RoundTrip(req)
+	if err != nil {
+		// Transport error AND no response: state stays as-is per definition B.
+		// The Proxy itself may have called observe(...) on a cache hit before
+		// returning an error path; we don't second-guess that.
+		if resp == nil {
+			h.reportError(fmt.Errorf("httptape: probe transport: %w", err))
+			return
+		}
+	}
```
tibtof, line 439: Non-blocking — minor: this swallows the case where `err != nil && resp != nil` (e.g. a partial body read). The Proxy in this repo doesn't return that combination, but `http.RoundTripper` allows it in general and `HealthMonitor` accepts arbitrary transports. Consider either reporting the error and returning, or adding a comment that the error is intentionally dropped when a response is also present.

health_test.go
```diff
@@ -0,0 +1,853 @@
...
+	}
+
+	if err := h.Close(); err != nil {
+		t.Fatalf("Close: %v", err)
+	}
+
+	// After close, no further calls within 3x the interval.
+	time.Sleep(80 * time.Millisecond)
+	if got := rt.callCount(); got > beforeClose {
+		t.Errorf("probe kept firing after Close: %d -> %d", beforeClose, got)
```
tibtof: Non-blocking — readability: this test demonstrably flakes (~20% without race, ~70% under `-race -count=20`) because of the underlying `runProbe` race I flagged in health.go. Once the implementation is fixed to deterministically stop probes after Close, this assertion (`got > beforeClose`) is the right one — keep it. If you ever loosen the contract to allow "at most one probe in-flight at the moment of Close," tighten the assertion to allow `+1` only when wg.Wait was racing with a tick. But the cleaner option is to fix the producer side.

tibtof (commented):

(Cannot submit a request-changes review on a self-authored PR; posting as a review comment instead.)

Excellent design and overall quality. ADR-28 is implemented faithfully: stdlib-only, no init(), no package-level state, errors wrapped with %w, table-driven tests where useful, godoc on every export, README + doc.go updated, CLI flags wired and guarded, backward-compat regression test in place, coverage of `health.go` averages 93%+ (most functions 100%). The hexagonal layering is preserved (no `internal/`, no new packages) and `go vet` is clean. `go.mod`/`go.sum` unchanged.

Two **blocking** concurrency issues plus a couple of non-blocking notes:

### Blocking #1 — `runProbe` does not honor Close deterministically
`TestHealthMonitor_CloseStopsProbe` is flaky: ~14/20 failures under `-race -count=20`, ~4/20 without race. Root cause: `select { case <-h.done: case <-ticker.C: }` picks at random when both are ready, so a probe can fire after `Close()` has been called (sometimes after `Close()` has already returned). The "no probes after Close" contract cannot hold today. See inline comment on health.go ~line 366 with two suggested fixes.

### Blocking #2 — send-on-closed-channel race in `subscribe()` and `applyObservation()`
Both register/snapshot under `h.mu` then send on subscriber channels *without* the lock. `Close()` closes those same channels under `h.mu`. The `select { case ch <- snap: default: }` pattern only protects against a full channel — sending on a closed channel panics. The window is small but reachable; it would crash the proxy. See inline comments on health.go lines 257 and 299. Cleanest fix: single-owner-of-close — only `unsub` (`dropSubscriber`) closes the channel; `Close()` signals via `h.done` and waits for handlers to unsubscribe themselves. Please also add a stress test that races subscribe/unsub against Close so `-race` will catch any regression.

### Non-blocking observations
- `runProbeOnce` (~line 399) silently swallows `err != nil && resp != nil` from arbitrary transports. Add a short comment or a `reportError` call.
- The inner watcher goroutine in `runProbe` is not tracked in `wg` — fine in practice (both exits are bounded) but a code comment would help future maintainers avoid adding work to it.
- `TestHealthMonitor_CloseStopsProbe`'s assertion is correct; the producer side is what needs fixing (see #1).

Once the two blocking items are fixed and `go test ./... -race -count=20` is clean for the health tests, this is good to ship.

tibtof:
## CHANGES_REQUESTED addressed

Three follow-up commits on `feat/121-health-endpoint`:

- `944c47f` fix(#121): re-check done channel in probe loop
- `380bdbd` fix(#121): single-owner-of-close discipline for SSE subscribers
- `97c6f88` test(#121): stress subscribe/observe/close concurrency

### Blocking #1 — `runProbe` race (health.go ~366)
Added the recommended non-blocking re-check of `<-h.done` inside the ticker case. When ticker and done are both ready, the recheck now wins and the loop returns without invoking `runProbeOnce`. Also tracked the inner watcher goroutine in `wg` so `Close()` waits for it (your second non-blocking note).

The test (`TestHealthMonitor_CloseStopsProbe`) had an additional race independent of the producer fix: the pre-Close sample races with an in-flight tick at the boundary (4-tick window over an 80ms sleep with a 20ms interval). Fixed by sampling the count *after* `Close()` returns — Close drains in-flight probes via `wg.Wait`, the recheck stops new ones, so the post-Close sample is deterministic. The contract assertion (`got > sample`) is unchanged. Verified clean under `go test -run TestHealthMonitor_CloseStopsProbe -race -count=200`.

### Blocking #2 — send-on-closed-channel race (health.go ~257, ~299)
Adopted the single-owner-of-close pattern you recommended: `dropSubscriber` remains the only place (besides Close itself) that closes a subscriber channel. Both producers — `applyObservation` broadcast and `subscribe` initial seed — now hold `h.mu` during the send. Since `dropSubscriber` and `Close` both also require `h.mu`, the send-on-closed window is closed by mutual exclusion. Overflowed subscribers are dropped in-place under the same lock. Subscribe also short-circuits a post-Close call by returning a closed channel + no-op unsub.

New stress test `TestHealthMonitor_SubscribeObserveCloseStress` runs 32 subscribers and 8 observers against a mid-flight `Close`. Plus `TestHealthMonitor_SubscribeAfterCloseReturnsClosedChannel` codifies the post-Close subscribe contract. Both clean under `-race -count=50`.

### Non-blocking #3 — `runProbeOnce` swallowed `(resp != nil, err != nil)` (~399)
Now closes `resp.Body` to avoid leaking the connection, surfaces the error via `reportError`, and returns. Comment in the source explains the http.RoundTripper contract.

### Non-blocking #4 — untracked watcher goroutine
Tracked in `wg` (see Blocking #1).

Full suite: `go test ./... -race -count=10` clean. `go vet` clean. `gofmt` clean.

tibtof:
## Reviewer verdict (round 2): APPROVED

(Cannot submit an approving review on a self-authored PR; posting as a comment instead.)

Both blocking issues from round 1 are correctly addressed and the new stress coverage is meaningful. **Ready to merge — no further rounds needed.**

### Blocking #1 — `runProbe` recheck-done — FIXED
Inner non-blocking re-check of `<-h.done` after the ticker case (health.go:397–402) closes the random-select window. Bonus: the inner watcher goroutine is now tracked in `wg` (health.go:378), satisfying the round-1 non-blocking note about untracked goroutines and ensuring `Close()` waits for it.

### Blocking #2 — single-owner-of-close — FIXED
The discipline is correctly enforced under `h.mu`:
- `applyObservation` holds `h.mu` for the full send via `defer Unlock()` (health.go:232–233).
- `broadcastLocked` schedules drops post-iteration to avoid map mutation during range, then closes only subscribers still in the map (health.go:256–271).
- `subscribe` holds `h.mu` across registration AND initial seed; post-Close short-circuit returns `(closed-ch, no-op-unsub)` so callers don't leak into the map and don't block (health.go:307–325).
- `dropSubscriber` checks membership before close — safe against double-close racing with `Close` or overflow drop (health.go:334–343).
- `serveStream` already handled `snap, ok := <-ch` correctly — closed channel from post-Close subscribe yields immediate clean return.

### Test correctness fix for `TestHealthMonitor_CloseStopsProbe` — VALID, not weakened
The pre-Close sample raced the in-flight tick at the measurement boundary itself — a test-side bug, not a missed contract violation. Sampling post-Close (where `wg.Wait` has drained in-flight probes and the recheck blocks new ones) measures exactly the contract: "no probes after Close returns". The assertion `got > afterClose` is preserved, so a real regression would still fail.

### `(resp != nil, err != nil)` handling — CORRECT
Body closed when present, error surfaced via `reportError`, return (health.go:429–438). Connection-leak-safe and consistent with the `reportError` pattern.

### Stress tests — MEANINGFUL
- `TestHealthMonitor_SubscribeObserveCloseStress`: 32 subscribers + 8 observers in tight loops for 50ms, then mid-flight `Close()`, then continued churn against a closed monitor. Verified clean under `go test -run TestHealthMonitor_(SubscribeObserveCloseStress|CloseStopsProbe) -race -count=200` (44s of pounding).
- `TestHealthMonitor_SubscribeAfterCloseReturnsClosedChannel`: codifies the post-Close subscribe contract.

### Verifications run
- `go test ./... -race -count=10` — clean (16s).
- `go test -run 'TestHealthMonitor_(SubscribeObserveCloseStress|SubscribeAfterCloseReturnsClosedChannel|CloseStopsProbe)' -race -count=50` — clean.
- `go test -run 'TestHealthMonitor_(SubscribeObserveCloseStress|CloseStopsProbe)' -race -count=200` — clean.
- `gofmt -l health.go health_test.go` — clean.
- `go vet ./...` — clean.
- Scope: only `health.go` and `health_test.go` touched between 70e4c58..97c6f88. `proxy.go`, `cmd/httptape/main.go`, `doc.go`, `README.md` confirmed unchanged.
- `TestProxy_HealthDisabledByDefault` still passes under `-race -count=5`.
- `wg` ordering for the inner `Add(1)` is safe (outer counter already 1 from `start()`); `waitForGoroutineCount` asserts no leaks.

Issue #128, opened 2026-04-16
chore: documentation drift cleanup (Go version, image size, missing flags/options/fields, example fixes)


tibtof:
## Problem statement

A docs audit against `main` @ `bf456dc` surfaced a cluster of small drift issues across the docs site, README, and LLM-targeted summaries. Each is small individually, but together they erode user trust (wrong Go version, wrong image size, missing public API surface, and example code that does not actually run as written).

## User story

As a Go developer evaluating or onboarding to httptape, I want the docs, README, and LLM-targeted summaries to accurately reflect the current code (correct Go version, correct image size, all CLI flags listed, all public options/fields documented, runnable examples), so that I can trust what I read and not get blocked by gaps or inconsistencies.

### 1. Go version

- [ ] `docs/getting-started.md:7` updated from \"Go 1.22 or later\" to \"Go 1.26 or later\".
- [ ] `docs/index.md:53` updated from \"Go 1.22 or later\" to \"Go 1.26 or later\".
- [ ] A grep across the repo confirms no other doc file claims a Go version below `1.26` for the runtime requirement (changelog/historical notes are fine to leave alone).

### 2. Docker image size

- [ ] The current Docker image is built locally and its size measured (e.g. `docker build` then `docker images tibtof/httptape:latest --format \"{{.Size}}\"`). The measurement command and resulting number are recorded in the PR description.
- [ ] The single, measured number replaces every claim of image size across:
  - `README.md:21`, `README.md:103`, `README.md:260` (currently \"6 MB\", including the comparison table)
  - `docs/docker.md:76` (currently \"~10 MB\")
  - `docs/ui-first-dev.md:384` (currently \"~10 MB\")
  - `llms.txt:104` (currently \"~10 MB\")
  - `llms-full.txt:104` (currently \"~10 MB\")
- [ ] The same number/format is used in all eight locations (no \"6 MB\" in one place and \"~10 MB\" in another after the change).
- [ ] The docs use language that does not promise an exact byte count going forward (e.g. \"~X MB\" rather than \"X.000 MB\") so future small drifts do not re-trigger this issue.

### 3. Missing CLI flags in `docs/cli.md`

- [ ] `record` flag table includes `--tls-cert`, `--tls-key`, `--tls-ca`, `--tls-insecure` (defined in `cmd/httptape/main.go:207-210`). Each entry links to `docs/tls.md` rather than duplicating the description.
[...]

PR #134, opened 2026-04-16
chore: enrich Docker Hub & GHCR registry pages — OCI labels, README sync, README rebalance (#122)

### LINKED ISSUE #122: chore: enrich Docker Hub & GHCR registry pages (OCI labels, README sync, README parity)
## Problem

Both of httptape's container registry pages currently underrepresent the project, and users who land on them from README badges learn nothing about what httptape is or how to use it.

- **Docker Hub** (https://hub.docker.com/r/tibtof/httptape) shows *No overview available* — the repository description is empty.
- **GHCR** (https://github.com/orgs/VibeWarden/packages/container/package/httptape) is similarly bare. GHCR auto-displays the source repo's README and adds a *Source* link to the package page, but only when the published image carries `org.opencontainers.image.source`. The current `Dockerfile` defines **zero OCI image labels**, so neither auto-link nor auto-README appears.
- The current `Dockerfile` (root of httptape repo) is bare metadata-wise — no `LABEL` instructions at all.
- The README mentions GHCR only once, on line 238 (\"Also available on GHCR: \`ghcr.io/vibewarden/httptape\`\"), buried at the end of the Docker section. GHCR is underrepresented relative to Docker Hub even though it is the primary, org-owned distribution channel.

## User story

As a Go developer evaluating httptape, when I follow a Docker Hub or GHCR badge from the README (or arrive via a search engine), I want the registry page to show a useful description, an obvious link back to the source repository, and the same getting-started snippets I would see in the README, so that I can decide whether to pull the image without having to chase links elsewhere.

### 1. OCI image labels in the Dockerfile

- [ ] The root `Dockerfile` declares (at minimum) the following OCI image labels on the final image:
  - [ ] `org.opencontainers.image.title` = `httptape`
  - [ ] `org.opencontainers.image.description` = a short tagline describing the image (one sentence, registry-page friendly)
  - [ ] `org.opencontainers.image.source` = `https://github.com/VibeWarden/httptape`
  - [ ] `org.opencontainers.image.url` = `https://github.com/VibeWarden/httptape` (or the docs URL — architect's call)
  - [ ] `org.opencontainers.image.documentation` = the published docs URL (currently the GitHub Pages site served by `.github/workflows/docs.yml`)
  - [ ] `org.opencontainers.image.licenses` = `Apache-2.0`
  - [ ] `org.opencontainers.image.version` = populated **at build time** from the release tag, via a Docker build-arg wired into the existing `.github/workflows/docker.yml` build step. For non-tag builds (push to `main`, PRs) it is acceptable for `version` to fall through to a sensible default such as the short commit SHA or `dev` — final choice is the architect's.
- [ ] After the next tagged release, `docker inspect` on the published image (both registries, both architectures) shows all of the above labels populated.
- [ ] The existing `docker/metadata-action@v5` step in `.github/workflows/docker.yml` continues to work; any labels it emits and any labels declared in the Dockerfile coexist without conflict (Dockerfile labels are baked into the image; metadata-action labels are applied at build time — both are fine).

### 2. Docker Hub README sync on release

- [ ] `.github/workflows/` is **extended** (not replaced) with a step that publishes the GitHub README to Docker Hub as the repository description on every released image push.
- [ ] The de-facto standard `peter-evans/dockerhub-description` GitHub Action is the assumed mechanism. (This is a CI action, not a Go module dependency — it does **not** violate the stdlib-only rule from `CLAUDE.md`.)
- [ ] The step uses the existing `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` secrets (already present in the org secrets store, since `.github/workflows/docker.yml` already pushes to Docker Hub with them). The architect should confirm scope of the existing token is sufficient for description writes; if not, the spec must call out the need for a new token with the right scope.
- [ ] The step runs only on tag pushes (`v*`), not on every push to `main`, to avoid noise.
- [ ] The short tagline used in the Docker Hub *short description* field (separate from the long README body) is set explicitly and matches the value used in `org.opencontainers.image.description`.
- [ ] After the next tagged release, the Docker Hub page at https://hub.docker.com/r/tibtof/httptape shows the project README rendered as the long description, and a non-empty short description.

### 3. README parity for GHCR

- [ ] The `## Docker` section of the root `README.md` is updated so that **GHCR is given equal billing with Docker Hub at the top of the section**, not relegated to a one-liner footnote at the bottom.
- [ ] Both registry image references appear next to each other in the opening of the section (e.g. listed as the two ways to pull the image), with a brief note that they are equivalent and updated in lockstep by the same release workflow.
- [ ] The existing `docker run` examples need not be duplicated for both registries; one canonical set of examples is fine, but the chosen registry name in those examples should not implicitly endorse one over the other (architect's call on which to use in the runnable snippets).

### 4. GHCR auto-link verification

- [ ] After the next tagged release ships an image with `org.opencontainers.image.source` set, the GHCR package page (https://github.com/orgs/VibeWarden/packages/container/package/httptape) shows:
  - [ ] A *Source* link pointing back to `VibeWarden/httptape`.
  - [ ] The repository README rendered on the package overview.
- [ ] This is a **post-release verification step**, not a build-time test. The acceptance criterion is met by a manual confirmation comment on the issue/PR after the next tagged release lands.

## Out of scope (explicit non-goals)

- **Image signing / cosign / keyless signing.** Worth doing later, but a separate concern with its own threat model and key-management story. Not in this issue unless trivially included by the same workflow.
- **SBOM generation / attestations** (`docker buildx build --sbom=true`, `--provenance=true`). Same reasoning — separate issue.
- **Vulnerability scanning** (Trivy, Grype, Snyk) wired into the release workflow. Separate issue.
- **Adding new GitHub secrets**. The spec assumes the existing `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` secrets are sufficient. If they are not, the architect should flag this back to PM rather than silently adding new secret requirements.
- **Renaming or moving the Docker Hub repository** (e.g. from `tibtof/httptape` to a `vibewarden/httptape` org account). Out of scope here even though it would be a natural follow-up.
- **Versioned docs / per-tag READMEs on the registry pages.** The Docker Hub long description always reflects the README from the most recent tagged release. Multi-version docs are not in scope.
- **Changing the Dockerfile build process, base image, or final image size.** Labels are metadata only; they must not change layer count meaningfully (a single consolidated `LABEL` instruction is fine; many separate `LABEL` instructions are also fine — architect's call).

## Constraints

- **Extend, do not replace** the existing release CI. The relevant workflows are `.github/workflows/docker.yml` (build + push to GHCR + Docker Hub, multi-arch amd64+arm64) and `.github/workflows/release.yml` (tests + GitHub Release notes). Either is a valid host for the README-sync step; the architect should decide which is the more cohesive home and justify the choice.
- **Stdlib-only constraint applies to the Go module, not to CI.** GitHub Actions used in CI workflows are not Go dependencies and do not count against the v1 stdlib-only rule.
- **Docker Hub README sync only takes effect on the *next* push** (whether that is a release tag or a manual workflow_dispatch run). Merging this issue's PR does not retroactively update the registry page — the architect/dev should call this out in PR description so reviewers know not to look for an immediate visible change on hub.docker.com.
- The OCI label `org.opencontainers.image.source` must point at the **canonical org repo** (`VibeWarden/httptape`), not at any personal fork, so that GHCR's auto-link resolves to the right place.

## Open questions (for the architect)

- Should the Dockerfile use a single consolidated `LABEL` instruction (smaller image, slightly less readable) or one `LABEL` per key (more readable, marginally more layers — note: with BuildKit, `LABEL`s do not add layers in practice)?
- Should the README-sync step live in `release.yml` (logically grouped with \"this is what we publish on a release\") or in `docker.yml` (logically grouped with \"everything that talks to the registries\")? Either is defensible.
- Is the existing `DOCKERHUB_TOKEN` scoped wide enough to write the repository description, or does it only have push scope? The architect should verify before designing around it.

## Deliverables expected from downstream stages

- **Architect**: a short design doc (this is near-trivial scope, so brevity is fine) covering: exact list of `LABEL` lines and how `version` is wired from build-arg, which workflow file hosts the README sync, the exact action version pin, and the README diff sketch for the Docker section.
- **Dev**: implementation in a single PR.
- **Reviewer**: standard review, plus a manual post-release verification of the GHCR and Docker Hub pages once the next tag is cut (this is the only acceptance criterion that cannot be confirmed pre-merge).

tibtof:
Closes #122. Implements ADR-30 in `decisions.md`.

## Summary

Three logical commits, single PR — pure docs/CI change, zero Go code touched:

1. **`Dockerfile`** — adds `ARG VERSION=dev` and a single consolidated `LABEL` block in the final stage declaring 7 `org.opencontainers.image.*` keys (`title`, `description`, `source`, `url`, `documentation`, `licenses`, `version`). These are the floor for local/downstream builds; `docker/metadata-action@v5` overrides overlapping keys on CI-published images with richer values (plus emits `revision` and `created`).
2. **`.github/workflows/docker.yml`** — two surgical edits:
   - `build-args: VERSION=<tag-or-dev>` on the existing `Build and push` step so the Dockerfile-baked `version` label resolves to the semver tag on tag pushes.
   - New `Update Docker Hub description` step using `peter-evans/dockerhub-description@v5`, gated on `startsWith(github.ref, 'refs/tags/v')`, syncing `README.md` as the long description and a fixed 98-char tagline as the short description.
3. **`README.md`** — rebalances the `## Docker` section so GHCR has equal billing at the top of the section (two-line `docker pull` block + one-line orientation note). Runnable `docker run` examples stay on `tibtof/httptape` to match the rest of the docs surface; the docs-wide image-name rebalance is owned by #128.

## Pre-flight verification

- **`https://vibewarden.dev/docs/httptape/`** → returns **200** (verified via `curl -I`); safe to use as the `org.opencontainers.image.documentation` label value.
- **Local Dockerfile build (default)** → `docker build -t httptape:label-check .` succeeds; `docker inspect ... --format '{{json .Config.Labels}}'` shows all 7 `org.opencontainers.image.*` keys populated, `version=dev`.
- **Local Dockerfile build (build-arg)** → `docker build --build-arg VERSION=v0.9.0-rc1 -t httptape:label-check-v .` succeeds; `org.opencontainers.image.version=v0.9.0-rc1` confirmed via `docker inspect`.
- **Workflow YAML lint** → `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/docker.yml'))"` → OK. All other workflow files (`docs.yml`, `release.yml`, `scorecard.yml`, `test.yml`) also lint clean.
- **Tagline length** → `printf '%s' "<tagline>" | wc -m` → **98 characters** (well under Docker Hub's 100-char short-description limit).
- **Go build / vet / test** → `go build ./...`, `go vet ./...`, `go test ./... -race -count=1` all clean (no Go code modified, but verified as a no-regression check). The new `test.yml` workflow gate will exercise the same on PR.

## Verify-step the maintainer must run before tagging v0.9.0

ADR-30 calls out a pre-merge token-scope verification. I cannot inspect the Docker Hub PAT scope from this environment — that's a Hub-UI / user-managed action. **Please verify before merging or before cutting v0.9.0:**

1. Visit Docker Hub → Account Settings → Personal Access Tokens.
2. Find the token wired into `DOCKERHUB_TOKEN` (created `2026-04-02T19:20:16Z` per `gh secret list`).
3. Required scope: **`Repository: Read, Write, Delete`** (or whatever Hub's current label is for "edit repository description"). The newer-style scopes that include "Public Repo Read, Read & Write" with description-edit permission also work.
4. **`Read & Write` (push-only) is insufficient** — the action calls `PATCH /v2/repositories/{namespace}/{repository}/` and will fail silently (step exits 0 in some action versions) if scope is too narrow.
5. If insufficient: rotate the token to the wider scope, update the `DOCKERHUB_TOKEN` secret in repo settings, then merge. If sufficient: merge as-is.

The silent-failure mode if not caught: the v0.9.0 release push runs the workflow, the `Update Docker Hub description` step exits green, but <https://hub.docker.com/r/tibtof/httptape> still shows "No overview available". Pre-flight verification avoids this round trip.

## Reviewer notes

- **No immediate visible change on Docker Hub** after merging this PR. The `Update Docker Hub description` step is gated on tag pushes only, so the Hub overview page will not update until the next `v*` tag (intended: **v0.9.0**) is cut.
- **GHCR auto-link verification is post-release**. Acceptance criterion 4 requires a manual check after v0.9.0 ships that the GHCR package page shows a Source link back to `VibeWarden/httptape` and renders the README. The 8-step protocol is in ADR-30 and will be performed by the reviewer/PM after the tag.
- **Pre-merge timing matters.** This PR must merge before v0.9.0 is tagged so the v0.9.0 image carries the labels and the Hub README sync fires on the v0.9.0 push. If we ship v0.9.0 without it, the user-visible benefits slip to v0.9.1+.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./... -race -count=1` clean
- [x] `docker build .` produces image with all 7 OCI labels populated
- [x] `docker build --build-arg VERSION=...` overrides the `version` label
- [x] `docker.yml` parses as valid YAML
- [x] All other workflow YAML files still parse
- [x] `https://vibewarden.dev/docs/httptape/` returns 200
- [x] Tagline ≤ 100 characters
- [ ] Maintainer verifies `DOCKERHUB_TOKEN` scope per protocol above
- [ ] Post-v0.9.0: 8-step verification of Hub overview, GHCR Source link, GHCR README render, per-arch label set (out-of-scope for this PR; tracked on #122 until the verifier comments)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -68,5 +68,18 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            VERSION=${{ github.ref_type == 'tag' && github.ref_name || 'dev' }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+      - name: Update Docker Hub description
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
+        uses: peter-evans/dockerhub-description@v5
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          repository: tibtof/httptape
+          short-description: "HTTP traffic recording, redaction, and replay — embeddable Go library, CLI, and 6 MB Docker image."
+          readme-filepath: ./README.md
+          enable-url-completion: true
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,10 @@ RUN CGO_ENABLED=0 GOOS=linux go build \
 # --- Final stage ---
 FROM scratch
 
+# Build-arg (overridden by CI on tag pushes; defaults to "dev" for local
+# builds and non-tag CI runs).
+ARG VERSION=dev
+
 # Import CA certs so record mode can dial HTTPS upstreams.
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 
@@ -31,5 +35,18 @@ VOLUME ["/fixtures", "/config"]
 EXPOSE 8081
 
 USER 65534
+
+# OCI image labels. See https://github.com/opencontainers/image-spec/blob/main/annotations.md
+# Note: docker/metadata-action@v5 in CI also emits labels; for overlapping
+# keys, the metadata-action value overrides on the published image. These
+# Dockerfile labels are the floor that guarantees metadata exists on local
+# `docker build .` and on downstream re-builds.
+LABEL org.opencontainers.image.title="httptape" \
+      org.opencontainers.image.description="HTTP traffic recording, redaction, and replay — embeddable Go library, CLI, and 6 MB Docker image." \
+      org.opencontainers.image.source="https://github.com/VibeWarden/httptape" \
+      org.opencontainers.image.url="https://github.com/VibeWarden/httptape" \
+      org.opencontainers.image.documentation="https://vibewarden.dev/docs/httptape/" \
+      org.opencontainers.image.licenses="Apache-2.0" \
+      org.opencontainers.image.version="${VERSION}"
 
 ENTRYPOINT ["httptape"]
--- a/README.md
+++ b/README.md
@@ -247,6 +247,17 @@ httptape import  --fixtures ./mocks --input bundle.tar.gz
 
 ## Docker
 
+Pull from either registry — the same image is published to both on every
+release:
+
+```bash
+docker pull tibtof/httptape           # Docker Hub
+docker pull ghcr.io/vibewarden/httptape   # GHCR
+```
+
+Examples below use `tibtof/httptape`; substitute `ghcr.io/vibewarden/httptape`
+freely.
+
 ```bash
 # Replay mode
 docker run -v ./mocks:/fixtures -p 8081:8081 tibtof/httptape serve --fixtures /fixtures
@@ -261,8 +272,6 @@ docker run -v ./cache:/fixtures -v ./config.json:/config/config.json -p 8081:808
     tibtof/httptape proxy --upstream https://api.example.com \
     --fixtures /fixtures --config /config/config.json
 ```
-
-Also available on GHCR: `ghcr.io/vibewarden/httptape`
 
 ## Testcontainers
 


tibtof:
## Reviewer Agent: APPROVED

(Posted as a comment because GitHub blocks self-review on the same account; treat this as the equivalent of an approving review.)

Reviewed all three commits (8a2ce93, 01277b9, ce33adf) against ADR-30. The implementation matches the normative skeleton verbatim.

**Verified:**
- **Dockerfile**: `ARG VERSION=dev` + single consolidated `LABEL` block in the final stage with all 7 `org.opencontainers.image.*` keys (title, description, source, url, documentation, licenses, version). Label values match ADR-30 exactly. `documentation` URL `https://vibewarden.dev/docs/httptape/` returns 200.
- **`.github/workflows/docker.yml`**: `build-args: VERSION=${{ github.ref_type == 'tag' && github.ref_name || 'dev' }}` wired correctly — resolves to the tag (e.g. `v0.9.0`) on tag push, `dev` otherwise. New `Update Docker Hub description` step uses `peter-evans/dockerhub-description@v5`, gated on `github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')` (correct: skips PR + main pushes), reuses existing `DOCKERHUB_USERNAME`/`DOCKERHUB_TOKEN`. YAML parses cleanly. `short-description` is exactly 100 chars (at Hub's hard limit by design) and matches the Dockerfile `description` label byte-for-byte.
- **README**: `## Docker` rebalance puts both registries side-by-side at the top with the "same image is published to both" note; runnable examples intentionally stay on `tibtof/httptape` per ADR-30 (docs-wide rebalance is owned by #128).
- **Hygiene**: No Go file, `go.mod`, `go.sum`, or other workflow/doc touched. `go test ./... -race` clean. `go vet` clean. (Pre-existing gofmt drift on `main` is unrelated to this PR.)
- **Label coexistence with `docker/metadata-action@v5`** is correctly documented inline in the Dockerfile comment block — Dockerfile labels are the floor, metadata-action overrides on overlapping keys, `documentation` is unique to the Dockerfile.

### Ready to merge — no further review rounds needed.

**Maintainer action item before merge / before v0.9.0 tag** (already flagged in the PR body, restating for the merge gate): please complete the 5-step `DOCKERHUB_TOKEN` scope verification per ADR-30 section "Token scope verification (pre-flight)". If the token is `Read & Write` (push-only), the new `Update Docker Hub description` step will succeed-but-no-op silently on the v0.9.0 release push, and the Hub overview won't update. Required scope is `Read, Write, Delete` on Repositories.

Acceptance criterion 4 (GHCR auto-link verification via the 8-step protocol) remains correctly out of scope for this PR — that is post-release, performed against the v0.9.0 published image. Issue #122 stays open until that verification passes.

Issue #46, opened 2026-03-30
Stateful scenarios (state machine)


tibtof:
## Problem Statement

Some API workflows involve multi-step interactions where the response to a request depends on previous requests (e.g., create then fetch, login then access protected resource). Static fixture matching cannot model these stateful flows.

## User Story

As a Go developer, I want to define stateful scenarios (state machines) for multi-step API interactions so that my mock server can simulate realistic workflow sequences like authentication flows or CRUD operations.

## Acceptance Criteria

- [ ] State machine model: scenarios define states and transitions triggered by matching requests
- [ ] Each state can have its own set of fixture responses
- [ ] State transitions happen automatically when a matching request is received
- [ ] API to define scenarios programmatically
- [ ] Documentation and examples

## Out of Scope

- Persistent state across server restarts
- Visual state machine editor

PR #154, opened 2026-04-17
feat(#45): response templating -- Mustache-style request data interpolation in fixtures

### LINKED ISSUE #45: Response templating
## Problem Statement

Recorded fixture responses are static: the exact bytes captured during recording are replayed for every matching request. This forces users to create a separate fixture for every unique request variation (e.g., each distinct `Idempotency-Key`, each unique `X-Request-ID`). Response templating allows a single fixture to interpolate values from the incoming request at serve time, dramatically reducing fixture count and improving test realism.

## User Story

As a Go developer, I want response fixtures to interpolate values from the incoming request (headers, query params, body fields) at serve time, so that a single fixture can produce dynamic, request-aware responses without requiring one fixture per variation.

## Headline Example

A payment API fixture echoes back the request's `Idempotency-Key` header:

```json
{
  "response": {
    "status_code": 200,
    "headers": {
      "X-Idempotency-Key": ["{{request.headers.Idempotency-Key}}"]
    },
    "body": "{\"id\": \"pay_123\", \"idempotency_key\": \"{{request.headers.Idempotency-Key}}\"}"
  }
}
```

With static fixtures, each unique key requires its own fixture. With templating, one fixture handles all keys.

## Template Syntax

**Mustache-style double curly braces**: `{{request.<accessor>}}`

Rationale for Mustache over alternatives:
- Most widely recognized template delimiter across ecosystems (Mustache, Handlebars, Go templates, WireMock)
- Developers immediately understand what `{{...}}` means without documentation
- The `{{` sequence is extremely rare in real HTTP response bodies (JSON uses single braces)
- WireMock uses the same delimiters, easing migration for teams switching tools
- Square-bracket alternatives `[[ ]]` would create ambiguity with JSON arrays

## Available Data Accessors

| Expression | Description | Example value |
|---|---|---|
| `{{request.method}}` | HTTP method (uppercase) | `GET` |
| `{{request.path}}` | Full URL path | `/users/42/orders` |
| `{{request.url}}` | Full request URL (as stored) | `https://api.example.com/users?page=2` |
| `{{request.headers.<Name>}}` | Header value (canonical key, first value) | `{{request.headers.Idempotency-Key}}` |
| `{{request.query.<name>}}` | Query parameter (first value) | `{{request.query.page}}` |
| `{{request.body.<json-path>}}` | JSON body field using existing path syntax | `{{request.body.user.email}}` |

### Header name resolution
Header names use Go's canonical format (`http.CanonicalHeaderKey`). For example, `{{request.headers.Content-Type}}` matches the `Content-Type` header regardless of how the client cased it.

### Query parameter resolution
Query parameters are matched by exact key name. If a parameter appears multiple times, the first value is used.

### Body path syntax
Body paths reuse the same `$.`-prefixed JSONPath-like syntax already established by `RedactBodyPaths` and `FakeFields` in `sanitizer.go`. Inside a template expression, the `$.` prefix is omitted for readability:
- `{{request.body.user.email}}` refers to `$.user.email`
- `{{request.body.items[*].sku}}` is NOT supported (array wildcards produce multiple values; undefined behavior for string interpolation)

Only scalar leaf values (strings, numbers, booleans) are interpolated. Objects and arrays resolve to empty string (or error in strict mode).

## Where Templating Applies

- **Response body** (any Content-Type that is text-based: JSON, XML, HTML, plain text)
- **Response header values**
- **NOT** response status code (integer, not meaningful to template)
- **NOT** request-side fields (recording is write-time, not serve-time)

## When Templating Runs

At **serve time**, inside `Server.ServeHTTP`, after match resolution and delay, before writing the response. The recorded fixture on disk is never modified. Templating is a read-time transformation.

## Default Behavior

- **Enabled by default**. Rationale: `{{...}}` is vanishingly rare in real HTTP payloads. Users who hand-author fixtures expect templating to work without extra configuration. Users who only replay recorded traffic (no `{{` in fixtures) see zero behavioral change.
- `WithTemplating(false)` server option disables all template processing (skip the scan/replace step entirely).

## Error Handling for Unresolvable References

Two modes:

1. **Lenient (default)**: Unresolvable template expressions are replaced with an empty string. This is safe for production-like tests where a missing header is acceptable.
2. **Strict**: `WithStrictTemplating(true)` server option. Unresolvable expressions cause the server to return an error response (500) with an `X-Httptape-Error: template` header and a body describing which expression failed. This is useful in tests where the developer wants to catch fixture authoring mistakes.

## Acceptance Criteria

- [ ] Template expressions in response **bodies** are resolved against the incoming request at serve time
- [ ] Template expressions in response **header values** are resolved against the incoming request at serve time
- [ ] `{{request.method}}` resolves to the HTTP method string (e.g., `GET`, `POST`)
- [ ] `{{request.path}}` resolves to the URL path of the incoming request
- [ ] `{{request.url}}` resolves to the full request URL string
- [ ] `{{request.headers.<Name>}}` resolves to the first value of the named header (canonical key match, case-insensitive lookup)
- [ ] `{{request.query.<name>}}` resolves to the first value of the named query parameter
- [ ] `{{request.body.<path>}}` resolves scalar JSON body fields using the existing JSONPath-like syntax from `sanitizer.go` (dot-notation, no `$.` prefix in the expression)
- [ ] Non-scalar body values (objects, arrays) resolve to empty string in lenient mode
- [ ] Responses with no `{{` sequences incur negligible overhead (fast-path: skip processing)
- [ ] Templating is enabled by default on `NewServer`
- [ ] `WithTemplating(false)` server option disables all template processing
- [ ] `WithStrictTemplating(true)` server option causes unresolvable expressions to return HTTP 500 with `X-Httptape-Error: template` header
- [ ] In lenient mode (default), unresolvable expressions are replaced with empty string
- [ ] Templating does NOT modify the stored fixture (pure serve-time transformation)
- [ ] Response status code is never templated
- [ ] New code lives in `templating.go` and `templating_test.go`
- [ ] Integration point is in `server.go` only (no changes to recorder, sanitizer, matcher, or store)
- [ ] All new public types and functions have godoc comments
- [ ] Functional options pattern used for new server options
- [ ] No external dependencies (stdlib only)
- [ ] No `panic` in hot paths
- [ ] Minimum 90% test coverage on new code
- [ ] Race-clean under `go test -race`
- [ ] Table-driven tests preferred

## Out of Scope

These items are explicitly excluded from this story:

- **Conditionals** (`{{if ...}}`, `{{else}}`) -- not a logic engine
- **Loops** (`{{range ...}}`) -- not a logic engine
- **Custom helpers / filters** (e.g., `{{uppercase request.headers.X-Foo}}`) -- keep the expression language trivial
- **WireMock helper functions** (`{{randomValue}}`, `{{now}}`, `{{math}}`) -- not in scope
- **Path parameter extraction** (`{{request.path.id}}` from `/users/{id}`) -- requires path-template matching in the Matcher layer, which does not exist today (see Pre-flight Notes below). This is a separate story.
- **Array wildcard body paths** (`{{request.body.items[*].sku}}`) -- undefined how to interpolate multiple values into a single string position
- **Response status code templating** -- status is an integer, not meaningful to interpolate
- **Request-side templating** -- recording is write-time
- **Stateful scenarios** -- covered by #46
- **Config file support for templating options** -- can be added later to `config.go`

### 1. Path parameter extraction is NOT possible today

No existing matcher supports path templates (e.g., `/users/{id}`). `MatchPath()` does exact string comparison. `MatchPathRegex()` uses regex but does not extract named capture groups. Therefore, `{{request.path.<param>}}` (accessing a named path segment) cannot be implemented without first adding a path-template matcher that extracts named parameters and makes them available to the server.

**Recommendation**: Exclude `{{request.path.<param>}}` from this story. Keep `{{request.path}}` (full path string). Path parameter extraction should be a separate prerequisite story that adds a path-template matcher with named group extraction. This story can be extended later to consume those extracted parameters.

### 2. JSON-path engine exists and should be reused

`sanitizer.go` contains `parsePath()` and the `segment`/`parsedPath` types. `matcher.go` contains `extractAtPath()` which traverses a parsed JSON structure. Both are unexported but available within the package. These should be reused for `{{request.body.<path>}}` resolution rather than building a new parser.

### 3. Implementation approach (architect decision)

The architect should decide between:
- **Hand-rolled `{{...}}` scanner/replacer**: Simple string scan for `{{`, find matching `}}`, resolve the expression. No parsing of pipes, conditionals, or anything beyond the accessor syntax.
- **stdlib `text/template` with restrictions**: Parse the template, then walk the AST to reject any node that is not a simple field access (reject `ActionNode` with pipes, `IfNode`, `RangeNode`, etc.). More code for the restriction logic, but gets escaping and edge cases for free.

This is an architecture decision, not a PM decision. Both approaches satisfy the spec.

### 4. Integration point in ServeHTTP

Templating should run after match resolution, after delay, and before response write. In the current `ServeHTTP` pipeline, this is between the current step 7 (delay) and step 8 (write response). The matched `Tape`'s response body and headers are processed through the template engine, and the resolved values are written to the `http.ResponseWriter`.

## Dependencies

- No dependency on other open stories
- Future story for path-template matching (not yet filed) would extend this feature with `{{request.path.<param>}}` support
- #46 (Stateful scenarios) is independent but may eventually combine with templating

## Milestone

`milestone:future`

tibtof:
Closes #45

## Summary

- Adds response templating support that resolves `{{request.*}}` Mustache-style expressions in response bodies and headers against the incoming HTTP request at serve time
- Hand-rolled `{{...}}` scanner (~80 lines of Go, single `bytes.Index` loop) -- no `text/template` dependency
- Two exported resolver functions: `ResolveTemplateBody` and `ResolveTemplateHeaders` -- callable from `Server.ServeHTTP` and by advanced users composing custom servers
- Two new `ServerOption` functions: `WithTemplating(bool)` and `WithStrictTemplating(bool)`
- Reuses `parsePath()` from `sanitizer.go` and `extractAtPath()` from `matcher.go` for JSON body field resolution
- Templating enabled by default; fast-path `bytes.Contains` skip for bodies without `{{`

### Supported expressions

| Expression | Example |
|---|---|
| `{{request.method}}` | `GET`, `POST` |
| `{{request.path}}` | `/users/42/orders` |
| `{{request.url}}` | `/users?page=2` |
| `{{request.headers.<Name>}}` | Header value (case-insensitive via `http.CanonicalHeaderKey`) |
| `{{request.query.<name>}}` | First value of query parameter |
| `{{request.body.<json.path>}}` | Scalar JSON field (dots = path segments) |

### Example: idempotency key echo

```json
{
  "response": {
    "status_code": 200,
    "headers": {
      "X-Idempotency-Key": ["{{request.headers.Idempotency-Key}}"]
    },
    "body": "{\"id\": \"pay_123\", \"idempotency_key\": \"{{request.headers.Idempotency-Key}}\"}"
  }
}
```

## Files changed

- **`templating.go`** (new) -- scanner, resolver, `ResolveTemplateBody`, `ResolveTemplateHeaders`, `scalarToString`
- **`templating_test.go`** (new) -- 30+ table-driven test cases covering all expression types, edge cases, strict/lenient modes
- **`server.go`** (modified) -- 2 new fields (`templating`, `strictTemplating`), 2 new option functions, resolution hook in `ServeHTTP` step 8 (between delay and response write)
- **`server_test.go`** (modified) -- 14 new server-level integration tests for templating

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./... -race -count=1` passes (all tests green, race-clean)
- [x] `go vet ./...` passes
- [x] `gofmt` clean
- [x] Overall coverage 93.9%, templating.go functions 85-100% per function (target: 90%)
- [x] All 23 acceptance criteria from issue #45 addressed
- [x] No external dependencies (stdlib only)
- [x] No panics in hot paths
- [x] Stored fixtures not modified (pure serve-time transformation)
- [x] Non-`request.` namespaces left as literal text (forward-compat for `state.*` in #46)
- [x] `{{request.path.<param>}}` treated as unresolvable (forward-compat for future path-template matcher)

### DIFF
--- a/server.go
+++ b/server.go
@@ -14,16 +14,18 @@ import (
 // Server is safe for concurrent use by multiple goroutines. All fields are
 // immutable after construction.
 type Server struct {
-	store          Store
-	matcher        Matcher
-	fallbackStatus int              // HTTP status when no tape matches
-	fallbackBody   []byte           // response body when no tape matches
-	onNoMatch      func(*http.Request) // optional callback when no tape matches
-	cors           bool             // if true, add CORS headers to all responses
-	delay          time.Duration    // fixed delay before every response; zero means no delay
-	errorRate      float64          // fraction of requests that return 500 (0.0-1.0)
-	randFloat      func() float64   // random number generator (injectable for testing)
-	replayHeaders  map[string]string // headers injected into every replayed response
+	store            Store
+	matcher          Matcher
+	fallbackStatus   int                 // HTTP status when no tape matches
+	fallbackBody     []byte              // response body when no tape matches
+	onNoMatch        func(*http.Request) // optional callback when no tape matches
+	cors             bool                // if true, add CORS headers to all responses
+	delay            time.Duration       // fixed delay before every response; zero means no delay
+	errorRate        float64             // fraction of requests that return 500 (0.0-1.0)
+	randFloat        func() float64      // random number generator (injectable for testing)
+	replayHeaders    map[string]string   // headers injected into every replayed response
+	templating       bool                // if true, resolve {{request.*}} in responses
+	strictTemplating bool                // if true, unresolvable expressions produce 500
 }
 
 // ServerOption configures a Server.
@@ -109,6 +111,41 @@ func WithReplayHeaders(key, value string) ServerOption {
 	}
 }
 
+// WithTemplating controls whether Mustache-style {{request.*}} template
+// expressions in response bodies and headers are resolved against the
+// incoming request at serve time. When enabled, template expressions are
+// replaced with values from the request (method, path, headers, query
+// params, body fields). When disabled, no scanning or replacement occurs
+// and response bodies/headers are written exactly as stored.
+//
+// Templating is enabled by default. The {{...}} delimiter is vanishingly
+// rare in real HTTP payloads, so enabling it by default is safe. Users who
+// only replay recorded traffic (no {{ in fixtures) see zero behavioral
+// change, as the fast path skips processing for bodies without "{{".
+//
+// See also WithStrictTemplating for error handling of unresolvable
+// expressions.
+func WithTemplating(enabled bool) ServerOption {
+	return func(s *Server) { s.templating = enabled }
+}
+
+// WithStrictTemplating controls the error behavior when a template
+// expression cannot be resolved. When strict is true, an unresolvable
+// expression (e.g., referencing a missing header) causes the server to
+// return HTTP 500 with an X-Httptape-Error: template header and a body
+// describing which expression failed. When strict is false (the default),
+// unresolvable expressions are silently replaced with an empty string.
+//
+// Strict mode is useful in tests where fixture authoring mistakes should
+// be caught early. Lenient mode is useful for production-like tests where
+// a missing value is acceptable.
+//
+// Strict mode has no effect when templating is disabled via
+// WithTemplating(false).
+func WithStrictTemplating(strict bool) ServerOption {
+	return func(s *Server) { s.strictTemplating = strict }
+}
+
 // withRandFloat overrides the random number generator for testing.
 // This is unexported -- only used in tests to make error simulation
 // deterministic.
@@ -126,6 +163,7 @@ func withRandFloat(fn func() float64) ServerOption {
 //   - CORS disabled
 //   - no delay
 //   - no error simulation
+//   - templating enabled (lenient mode)
 //
 // The store must not be nil. Passing a nil store is a programming error and
 // will panic.
@@ -139,6 +177,7 @@ func NewServer(store Store, opts ...ServerOption) *Server {
 		matcher:        DefaultMatcher(),
 		fallbackStatus: http.StatusNotFound,
 		fallbackBody:   []byte("httptape: no matching tape found"),
+		templating:     true,
 	}
 	for _, opt := range opts {
 		opt(s)
@@ -162,8 +201,9 @@ func NewServer(store Store, opts ...ServerOption) *Server {
 //  3. Normal replay -- existing match-and-respond logic
 //  4. Per-fixture error override -- after matching, before writing
 //  5. Delay -- sleep before writing the real response
-//  6. Replay header injection -- override/add headers from WithReplayHeaders
-//  7. Write response
+//  6. Templating -- resolve {{request.*}} expressions in body and headers
+//  7. Replay header injection -- override/add headers from WithReplayHeaders
+//  8. Write response
 //
 // Performance note: ServeHTTP calls Store.List with an empty filter on every
 // request, resulting in an O(n) scan over all tapes. This is acceptable for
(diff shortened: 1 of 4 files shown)


tibtof (commented):
**APPROVED** -- ready to merge.

Thorough review of templating implementation for issue #45.

**What was reviewed:**
- `templating.go` (311 lines) -- hand-rolled `{{...}}` scanner using `bytes.Index` loops, two exported resolver functions, `readRequestBody` with body restoration
- `templating_test.go` (851 lines) -- 24 test functions covering scanner edge cases, all accessor types, strict vs lenient mode, body preservation, fast paths, scalar conversion
- `server.go` changes -- 2 new fields (`templating`, `strictTemplating`), 2 functional options, default-on in constructor, hook between delay (step 7) and write (step 9)
- `server_test.go` changes -- 14 new integration tests covering end-to-end behavior

**Checklist results (all pass):**
- Architecture: single flat package, no sub-packages, no `text/template` import, hand-rolled scanner as specified
- Path reuse: `parsePath` from `sanitizer.go` and `extractAtPath` from `matcher.go` are called directly (no copy-paste)
- Header case-insensitivity: uses `http.CanonicalHeaderKey()` then `r.Header.Get()`
- Path params (`request.path.<param>`): correctly fall through as unresolvable
- Non-request namespaces: left as literal text (forward-compatible with `state.*`)
- Default-on, lenient by default: confirmed in constructor and tests
- Strict mode: 500 + `X-Httptape-Error: template` header on unresolvable
- Fast path: `bytes.Contains(body, []byte("{{"))` check skips all work when no delimiters present, verified by pointer-identity test
- Fixture immutability: original `Tape` body not mutated, confirmed by `TestServer_Templating_DoesNotModifyStoredFixture`
- Request body preservation: `readRequestBody` restores via `io.NopCloser(bytes.NewReader(body))`
- No panics, no `init()`, no global state, no external dependencies
- `go test ./... -race -count=1` passes clean
- `go vet` clean
- `gofmt` clean on new code (pre-existing formatting issues in `server_test.go` are out of scope)
- `go.mod` / `go.sum` unchanged
- Coverage: `templating.go` functions range 85.7%-100%, overall package at 93.9%
- All exported symbols have godoc comments
- Errors wrapped with context (`fmt.Errorf("httptape: ...")`)
- No mocking frameworks, stdlib `testing` only

**Non-blocking observations (not requesting changes):**
1. Header/query resolution treats empty-string values as unresolvable (lines 90, 101 of `templating.go`). In practice this is fine since empty header values are extremely rare, but worth documenting if users ask.
2. `resolveTemplateString` line 264 allocates `[]byte(s)` for `bytes.Contains` where `strings.Contains(s, "{{")` would avoid the allocation. Negligible in practice.
3. Request body is read/restored twice when both body and header templates reference `request.body.*` (once in `ResolveTemplateBody`, once in `ResolveTemplateHeaders`). Functionally correct, minor inefficiency.
4. ADR-34 referenced in issue #45 comments was not written to `decisions.md`. The design decisions are documented in the issue comments instead. This is an upstream gap, not a PR deficiency.

PR #157, opened 2026-04-17
feat: Server-Sent Events support -- record, replay, proxy passthrough (#124)

### LINKED ISSUE #124: feat: Server-Sent Events (SSE) support — record, replay, proxy passthrough
## Problem statement

httptape's request/response model handles bodies as a single buffered blob. Server-Sent Events (SSE) responses are long-lived `text/event-stream` HTTP responses that emit a sequence of discrete events over time. Today, recording one of these against the existing pipeline either hangs (waiting for an EOF that may not come for minutes), buffers the entire stream into memory and loses inter-event timing, or both. Replaying one is not possible at all — the server hands the client a single static body with no flushing semantics. The dominant use case driving this is recording an LLM streaming completion (Claude / OpenAI / Anthropic / local LLM) once and replaying it deterministically in tests, but the same shape covers progress streams, notification feeds, and real-time dashboards.

This issue adds SSE as a first-class shape inside httptape's existing HTTP-shaped model. It is the SSE-shaped successor to the WebSocket issue closed in #49 — SSE is just HTTP, fits the existing tape/recorder/server/proxy types with a per-event extension rather than a parallel type system, and covers the demand that drove most of the WebSocket interest in the first place.

## User story

As a Go developer building or testing an application that calls a streaming HTTP API (most prominently an LLM streaming endpoint), I want httptape to record the SSE stream once, sanitize it on write, and replay it deterministically — with realistic, accelerated, or instant inter-event timing — so that I can test my streaming consumer against a stable fixture without hitting the live upstream on every run, without leaking secrets into the fixture, and without rewriting my consumer code to special-case SSE.

## Scope

Three modes — recording, replay, proxy passthrough — plus per-event sanitization on write. New behavior is additive: existing tapes, existing API surface, and non-SSE traffic must be byte-for-byte unaffected.

### 1. Detection
- The recorder identifies a response as SSE when its `Content-Type` is `text/event-stream` (case-insensitive on the type/subtype, parameter-tolerant, e.g. `text/event-stream; charset=utf-8`).
- Once detected, the recorder switches from buffer-the-body to event-streaming capture mode for the lifetime of that response.

### 2. Recording
- The original request is captured exactly as today.
- The response body is consumed as a stream and parsed against the SSE protocol: lines terminated by `\n`, `\r\n`, or `\r`; field-name/value separated by the first `:`; supported field names `event`, `data`, `id`, `retry`; lines beginning with `:` are comments and ignored; an event is dispatched on the first blank line; `data` lines accumulate and are joined by `\n` per the spec.
- Each dispatched event is persisted with its offset (in milliseconds) from the moment the response started. The recorded shape includes at least: `offset_ms`, `event` (omitempty), `data` (the joined payload), `id` (omitempty), `retry` (omitempty), and any future-proofing fields the architect deems necessary.
- The recorder must remain transparent to the caller: the original `http.Response.Body` continues delivering bytes to the calling code unchanged and on the original timing. Recording happens via tee/duplication, not by interposing a buffer that breaks streaming semantics for the consumer.
- An SSE response that is malformed mid-stream (invalid framing, bytes that don't parse as events) must not crash the recorder. The architect picks the recovery policy (truncate-and-mark, best-effort skip, fail-the-tape) and documents it.
- An SSE response that the upstream simply leaves open and never closes must not block other recorder work indefinitely. Behavior on upstream-disconnect mid-stream (cleanly persisted, partial flag, etc.) must be defined.

### 3. Tape JSON schema change (additive)
- The `RecordedResp` shape gains an optional list of recorded events. Field name is the architect's call; this spec uses `sse_events` as a working name.
- When `sse_events` is present, the tape represents an SSE response. When absent, the tape behaves exactly as today.
- Whether the existing `body` field stays empty for SSE tapes, holds the raw concatenated stream, or is left as the architect's call — must be settled in the design with a documented rationale. Backward-compatible reading of old tapes (no `sse_events` field) is required.
- `config.schema.json` must be updated to describe the new optional shape.

### 4. Replay (mock server)
- When a matched fixture has SSE events, the server responds with `Content-Type: text/event-stream`, sensible cache/connection headers (e.g. `Cache-Control: no-cache`), and uses `http.Flusher` to deliver each event before moving on.
- The server emits each event using SSE wire format (one or more `data:` lines, optional `event:`, `id:`, `retry:`, terminated by a blank line) reconstructed from the recorded fields.
- Inter-event timing is configurable per-server-instance via a functional option. Working name `WithSSETiming`, with three modes:
  - **realtime** (default) — replay the recorded `offset_ms` cadence as faithfully as scheduling allows.
  - **accelerated(factor)** — divide all intervals by `factor` (e.g. 10x for fast tests). Factor must be > 0.
  - **instant** — fire all events immediately back-to-back, no inter-event delay.
- Replay must respect client disconnect: when the client closes the connection mid-stream, the server stops emitting events for that connection promptly (no leaked goroutines, no pile-up).
- Multiple concurrent replays of the same SSE fixture must be supported without cross-talk and without serializing on a single shared resource.

### 5. Proxy passthrough
- In proxy mode, when the upstream response is SSE (detected as in section 1), the proxy streams events to the client as they arrive — no buffering of the whole stream before forwarding, since that defeats the point.
- In parallel, the proxy records the events into the configured store (L1 / L2 / both, per existing proxy semantics).
- The `X-Httptape-Source` header (`live` / `l1-cache` / `l2-cache`) is set on the response status line, before the first event flushes, exactly as for non-SSE responses today.
- L2 fallback for SSE: when the upstream is unreachable and a cached SSE fixture is available, the proxy replays it from cache using the configured replay timing (section 4). The decision of which timing applies in proxy fallback (default realtime, an option, etc.) is the architect's call but must be documented.
- When the upstream returns SSE, the proxy must not gzip/transform the stream in a way that breaks `Flusher` semantics or per-event delivery to the client.

### 6. Sanitization (on write)
- The existing `Sanitizer` pipeline must apply per-event to each `data` payload during recording, before the event is persisted. This is the same on-write principle httptape applies to non-SSE bodies — sensitive content must never reach disk.
- When `data` is JSON, body-path redaction rules apply to each event's payload as if it were an independent JSON body.
- When `data` is plain text, header-style redaction and pattern-based redaction must apply.
- Header sanitization for the response itself runs once, as today.
- Whether the existing `Sanitizer` interface and pipeline can apply per-event without extension, or whether it needs a small new entry point (e.g. a per-event hook), is **flagged as an open question** for the architect — the existing pipeline may already be reusable here, but a per-event view of the sanitization contract should be confirmed before implementation.
- Sanitization must not break the SSE wire format on replay: redacted `data` values must still serialize to a valid `data:` line (or multiple, if they contain newlines).

## Acceptance criteria

Behavior-based; each must be testable in isolation.

### Detection
- [ ] A response with `Content-Type: text/event-stream` triggers SSE capture mode.
- [ ] A response with `Content-Type: text/event-stream; charset=utf-8` (parameters present) also triggers SSE capture mode.
- [ ] Case differences in the type/subtype (e.g. `Text/Event-Stream`) do not prevent detection.
- [ ] A response without an SSE content type is unaffected by this change and is recorded exactly as today (verified by an existing-recorder regression test or an explicit non-SSE assertion).

### Recording
- [ ] A recorded SSE stream of N events produces a tape with N entries in `sse_events` (working name), in the order the upstream emitted them.
- [ ] Each recorded event preserves `data` exactly (after sanitization), preserves `event` / `id` / `retry` when the upstream sent them, and omits them when the upstream did not.
- [ ] Each recorded event has an `offset_ms` measured from the moment the response started; offsets are monotonically non-decreasing across the stream.
- [ ] SSE comment lines (lines beginning with `:`) are not persisted as events.
- [ ] Multi-line `data` is joined with `\n` per the SSE spec before persistence.
- [ ] The calling code's `http.Response.Body` reads see the original byte stream unchanged (verified by an integration test that consumes the body alongside recording and checks bytes are intact and arrive without spurious buffering).
- [ ] Concurrent recording of multiple SSE responses on different requests does not corrupt either tape and does not race (verified under `go test -race`).
- [ ] Upstream disconnect mid-stream produces a tape containing the events received so far (no panic, no hang, no goroutine leak).

### Replay (mock server)
- [ ] A tape with `sse_events` replays as a `text/event-stream` response with `http.Flusher` calls between events.
- [ ] With `WithSSETiming(SSETimingRealtime)` (default), the gap between events on the wire approximates the recorded `offset_ms` cadence within reasonable scheduling tolerance.
- [ ] With `WithSSETiming(SSETimingAccelerated(N))`, gaps are divided by `N` (within tolerance). `N <= 0` is rejected at option-construction time (or at server start) with a clear error.
- [ ] With `WithSSETiming(SSETimingInstant)`, all events flush back-to-back with no deliberate delay.
- [ ] When the client disconnects mid-stream, the replay handler stops emitting promptly and does not leak a goroutine (verified by a test that disconnects and asserts goroutine count returns to baseline).
- [ ] Two concurrent clients replaying the same SSE fixture each receive the full event sequence in order, without cross-talk.
- [ ] Reconstructed wire frames are valid SSE (each event terminated by a blank line; multi-line `data` re-emitted as multiple `data:` lines).

### Proxy passthrough
- [ ] When the upstream returns an SSE response, the proxy forwards events to the client as they arrive, not after the upstream closes the stream (verified by a test that asserts client-side receive timing tracks upstream emission, not stream end).
- [ ] In parallel, the proxy records the SSE events into the configured store; the resulting tape replays cleanly via the mock server.
- [ ] The `X-Httptape-Source` header is set correctly (`live` for upstream-served SSE; `l2-cache` for L2-fallback-served SSE) and is visible on the response before the first event flushes.
- [ ] When the upstream is unreachable and an L2 cached SSE fixture exists, the proxy serves it as a real SSE stream to the client (with timing per the configured replay policy) and sets the `X-Httptape-Source` header to `l2-cache`.

### Sanitization
- [ ] The configured `Sanitizer` runs on each event's `data` payload before persistence (verified by recording an event whose `data` contains a value matching a redaction rule and asserting the persisted tape has the redacted form).
- [ ] Redaction of JSON `data` payloads applies the same body-path rules used for non-SSE JSON bodies.
- [ ] Sanitized values still produce valid SSE wire frames on replay (no broken framing introduced by redaction tokens that contain newlines, colons, or other SSE-significant characters).
- [ ] No raw, unsanitized event content ever reaches the store on the recording path (verified by an explicit on-disk inspection in a test).

### Tape format / backward compatibility
- [ ] An existing tape (no `sse_events` field) loads and replays exactly as before.
- [ ] A tape with both `body` and `sse_events` is handled per the design's documented rule (the architect picks; the rule must be testable).
- [ ] `config.schema.json` is updated to describe the new optional field; the schema validates a representative SSE tape and continues to validate all existing tapes.

### Concurrency, safety, embedding
- [ ] No package-level state, no `init()`, no side effects on import.
- [ ] No new external dependencies — stdlib only.
- [ ] Functional options pattern for any new public constructor inputs.
- [ ] No panics in the recorder, replay handler, or proxy passthrough path. Constructor-guard panics on nil required dependencies are acceptable per existing convention.
- [ ] All new code is race-clean under `go test -race ./...`.
- [ ] On `Close` / shutdown, all in-flight SSE recording, replay, and passthrough operations terminate within the existing graceful-shutdown window without leaking goroutines.

### Tests
- [ ] Unit tests live next to the code they test (e.g. `sse_test.go`), table-driven where natural.
- [ ] Coverage for new code meets the project's 90% target.
- [ ] At least one integration test exercises the full loop: record an SSE response from a `httptest.Server`, persist to the store, replay through the mock server, assert the replayed events match the recorded ones (modulo timing tolerance) and arrive with `Flusher`-driven delivery.
- [ ] At least one integration test exercises proxy passthrough: SSE upstream, proxy in front, client receives events live AND tape is recorded; then the same fixture is served from L2 with upstream offline.
- [ ] At least one race-focused test in `race_test.go` exercises concurrent SSE recording/replay.

### Documentation
- [ ] Package godoc covers the new SSE behavior on `Recorder`, `Server`, and `Proxy`, and the new functional option(s).
- [ ] The README's recorder/server/proxy sections briefly describe SSE support and the LLM-streaming use case.
- [ ] The new tape field is documented (godoc + `config.schema.json`).

## Hard constraints (from CLAUDE.md — must respect)

- **stdlib only.** SSE protocol parsing is straightforward with `bufio`. SSE wire serving uses `net/http` plus `http.Flusher`.
- **No package-level state, no `init()`, no side effects on import.**
- **Functional options pattern** for new public constructor inputs (e.g. `WithSSETiming`).
- **Hexagonal-by-convention.** Live in new files (working names `sse.go`, `sse_test.go`), with hooks added to `recorder.go`, `server.go`, `proxy.go` only as needed. Do not reorganize the package.
- **Sanitize-on-write.** Per-event sanitization happens during recording, never as a post-export pass.
- **Tape JSON schema change must be backward-compatible for readers** (old fixtures without the new field keep working).
- **No panics in hot paths.** Constructor-guard panics on nil dependencies are the only acceptable panic.
- **Tests next to code, table-driven where natural, race-clean, 90% coverage target.**
- **Embeddable first.** Behavior must be reachable from the library API without going through any future CLI.

## Out of scope (explicitly)

- **WebSocket.** Closed in #49 — bidirectional message streams are a separate type-system rework and are not revisited here.
- **HTTP/2 server push.** Different mechanism; tracked separately by #47.
- **Server-to-server SSE-style streaming over non-`text/event-stream` content types** (e.g. NDJSON, application-specific chunked formats). Detection is strictly content-type-driven in this issue.
- **Replay-side reconnection logic driven by `Last-Event-ID`** (resuming a replay from a specific event ID on client reconnect). Note as a future enhancement.
- **Recording-side rate limiting or back-pressure for very high-throughput streams.** Note as a future enhancement.
- **Changing existing matchers, the L1/L2 cache layout, the sanitizer's existing public surface (beyond any small additive per-event hook the architect deems necessary), or `X-Httptape-Source` semantics.**

## Interaction with in-flight work

- **#121** (technical health endpoint surface for proxy mode, currently in development) ships an SSE endpoint **on the proxy itself** (`/__httptape/health/stream`). That endpoint is internal control-plane traffic emitted by the proxy and is **not** the kind of upstream SSE traffic this issue is about. It does, however, prove out the basic SSE wire plumbing (Flusher, `data: ...\n\n` framing, multi-subscriber lifecycle). If the dev work for #121 produces a small reusable internal helper for SSE wire serving, this issue may consume it. The design for this issue should leave room for that consumption without depending on it — if #121's helper is internal-only or unsuitable, this issue must still ship a self-contained SSE replay/proxy capability.

- **Final field name** for the per-event list on `RecordedResp` (working name `sse_events`).
- **Field set per recorded event.** `offset_ms`, `event`, `data`, `id`, `retry` is the working set; confirm or extend (e.g. for future-proofing or for partial-frame metadata on upstream-disconnect).
- **Relationship between `body` and `sse_events`** when both are present, or whether `body` is left empty for SSE tapes by construction.
- **Upstream-disconnect mid-stream policy** for the recorder (clean persist of partial events, explicit `partial` flag, etc.).
- **Malformed-SSE-mid-stream policy** for the recorder (truncate-and-mark, best-effort skip, fail-the-tape).
- **Replay timing tolerance** — what "approximates the recorded cadence within reasonable scheduling tolerance" means in concrete numbers for the test suite.
- **Acceleration factor type** — int, float64, or a typed `SSEAccelerationFactor` value object. Validation of `<= 0`.
- **Default replay timing for L2 proxy fallback** (realtime is the natural default; confirm or expose as an option).
- **Sanitization extension surface** — confirm whether the existing `Sanitizer` pipeline applies per-event without modification, or whether a small additive entry point (per-event hook) is needed. Document the decision.
- **Reuse of any SSE wire helper from #121** — at design time, decide whether to depend on, fork from, or ignore whatever helper #121's PR lands.
- **Whether to expose the recording-side SSE behavior under any opt-out** (default-on detection vs. an option to disable SSE-mode recording entirely). Smallest-surface default preferred.

tibtof:
Closes #124

## Summary

- **New `SSEEvent` type** in `sse.go` -- per-event value type with `OffsetMS`, `Type`, `Data`, `ID`, `Retry` fields
- **New `SSETimingMode` sealed interface** with three constructors: `SSETimingRealtime()`, `SSETimingAccelerated(factor)`, `SSETimingInstant()`
- **Recording**: Recorder detects `text/event-stream` responses and uses `sseRecordingReader` (io.TeeReader + background parser) to deliver original bytes to caller unchanged while parsing events in parallel. Tape persisted on body close. Opt-out via `WithSSERecording(false)`
- **Replay**: Server detects SSE tapes via `IsSSE()` and streams events with configurable timing (`WithSSETiming`). Sets SSE headers, flushes after each event, respects context cancellation
- **Proxy passthrough**: SSE responses are streamed live to the client while recording events to L1/L2 in the background. L2 fallback synthesizes a streaming response via `io.Pipe` with configurable timing (default instant via `WithProxySSETiming`)
- **Sanitization**: `RedactSSEEventData(paths...)` and `FakeSSEEventData(seed, paths...)` compose into existing `Pipeline`, applying body-path operations to each event's `Data` field independently
- **Schema change**: `RecordedResp` gains `SSEEvents []SSEEvent` with `json:"sse_events,omitempty"` -- fully backward compatible
- **Refactored** `health.go` to use shared `writeSSEEvent` helper from `sse.go`
- **Updated** `deepCopyTape` to copy `SSEEvents` slice

## Test plan

- [x] SSEEvent JSON round-trip and omitempty behavior
- [x] `isSSEContentType` detection (case-insensitive, parameter-tolerant)
- [x] `parseSSEStream` -- single events, multi-line data, event/id/retry fields, comments, unknown fields, malformed lines, partial events at EOF, empty input, reader errors
- [x] `writeSSEEvent` -- all field combinations, multiline data, write errors
- [x] `sseRecordingReader` -- basic parsing, upstream disconnect mid-stream, empty stream, close idempotency, concurrent reads
- [x] SSE timing modes -- Realtime preserves inter-event gaps, Accelerated divides them, Instant fires immediately, negative gap clamping, fractional factors
- [x] Replay timing verification with ADR-35 tolerance (+/- 50ms or 20%)
- [x] Server SSE replay end-to-end with instant and realtime timing
- [x] Server non-SSE tapes unchanged after SSE changes
- [x] Recorder SSE recording integration (full HTTP round-trip)
- [x] Recorder SSE recording disabled via `WithSSERecording(false)`
- [x] Proxy SSE success path -- events recorded to L1/L2
- [x] Proxy L2 fallback -- cached SSE tape replayed with instant timing
- [x] Sanitization -- `RedactSSEEventData` and `FakeSSEEventData` with JSON and non-JSON data
- [x] Schema backward compat -- old tape JSON without `sse_events` field deserializes cleanly
- [x] `RecordedResp.IsSSE()` with nil, empty, and populated events
- [x] `deepCopyTape` copies SSEEvents without aliasing
- [x] Parse-write round-trip fidelity
- [x] `health.go` streaming still works after `writeSSEEvent` refactor
- [x] Race detector clean: `go test -race -count=10` passes
- [x] Integration tests in `integration_test.go` for SSE record-replay and proxy E2E
- [x] Race tests in `race_test.go` for concurrent SSE recording and replay

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3,6 +3,7 @@ package httptape
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -650,7 +651,9 @@ func TestIntegration_RecordReplay_PostWithBody(t *testing.T) {
 // flow using MemoryStore: an upstream SSE server is recorded through the
 // Recorder, then replayed by the Server with instant timing.
 func TestIntegration_SSE_RecordReplay_E2E(t *testing.T) {
-	// Start an upstream that serves SSE.
+	const eventCount = 5
+
+	// Start an upstream that serves SSE with valid JSON event data.
 	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		flusher, ok := w.(http.Flusher)
 		if !ok {
@@ -661,10 +664,8 @@ func TestIntegration_SSE_RecordReplay_E2E(t *testing.T) {
 		w.WriteHeader(200)
 		flusher.Flush()
 
-		for i := 0; i < 5; i++ {
-			body := strings.NewReader("")
-			_ = body
-			w.Write([]byte("data: {\"n\":" + strings.TrimRight(strings.Repeat("0", i+1), "0") + "}\n\n"))
+		for i := 0; i < eventCount; i++ {
+			fmt.Fprintf(w, "data: {\"n\":%d}\n\n", i)
 			flusher.Flush()
 		}
 	}))
@@ -682,9 +683,13 @@ func TestIntegration_SSE_RecordReplay_E2E(t *testing.T) {
 	origBody, _ := io.ReadAll(resp.Body)
 	resp.Body.Close()
 
-	// The caller should have received the raw SSE bytes.
-	if !strings.Contains(string(origBody), "data:") {
-		t.Error("caller should see raw SSE data")
+	// The caller should have received the raw SSE bytes with valid JSON payloads.
+	origStr := string(origBody)
+	for i := 0; i < eventCount; i++ {
+		want := fmt.Sprintf("data: {\"n\":%d}", i)
+		if !strings.Contains(origStr, want) {
+			t.Errorf("original body missing %q", want)
+		}
 	}
 
 	// Replay.
@@ -709,9 +714,13 @@ func TestIntegration_SSE_RecordReplay_E2E(t *testing.T) {
 		t.Errorf("Content-Type = %q, want text/event-stream", ct)
 	}
 
-	// The replayed body should contain SSE data lines.
-	if !strings.Contains(string(replayBody), "data:") {
-		t.Error("replayed response should contain SSE data")
+	// Verify each event's data was replayed with the correct JSON payload.
+	replayStr := string(replayBody)
+	for i := 0; i < eventCount; i++ {
+		want := fmt.Sprintf("data: {\"n\":%d}", i)
+		if !strings.Contains(replayStr, want) {
+			t.Errorf("replayed body missing %q; got:\n%s", want, replayStr)
+		}
 	}
 }
 
--- a/integration_test.go
+++ b/integration_test.go
@@ -2,11 +2,13 @@ package httptape
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 )
 
 // --------------------------------------------------------------------
@@ -641,5 +643,154 @@ func TestIntegration_RecordReplay_PostWithBody(t *testing.T) {
 	}
 	if string(replayBody) != string(origBody) {
 		t.Errorf("replay body = %q, want %q", string(replayBody), string(origBody))
+	}
+}
+
+// TestIntegration_SSE_RecordReplay_E2E tests the full SSE record-and-replay
+// flow using MemoryStore: an upstream SSE server is recorded through the
+// Recorder, then replayed by the Server with instant timing.
+func TestIntegration_SSE_RecordReplay_E2E(t *testing.T) {
+	// Start an upstream that serves SSE.
+	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		flusher, ok := w.(http.Flusher)
+		if !ok {
+			http.Error(w, "no flusher", 500)
+			return
+		}
+		w.Header().Set("Content-Type", "text/event-stream")
+		w.WriteHeader(200)
+		flusher.Flush()
+
+		for i := 0; i < 5; i++ {
+			body := strings.NewReader("")
+			_ = body
+			w.Write([]byte("data: {\"n\":" + strings.TrimRight(strings.Repeat("0", i+1), "0") + "}\n\n"))
+			flusher.Flush()
+		}
+	}))
+	defer upstream.Close()
+
+	store := NewMemoryStore()
+	rec := NewRecorder(store, WithAsync(false))
+	client := &http.Client{Transport: rec}
+
+	// Record.
+	resp, err := client.Get(upstream.URL + "/events")
+	if err != nil {
+		t.Fatalf("GET: %v", err)
+	}
+	origBody, _ := io.ReadAll(resp.Body)
+	resp.Body.Close()
+
+	// The caller should have received the raw SSE bytes.
+	if !strings.Contains(string(origBody), "data:") {
+		t.Error("caller should see raw SSE data")
+	}
+
+	// Replay.
+	srv := NewServer(store, WithSSETiming(SSETimingInstant()))
+	replayTS := httptest.NewServer(srv)
+	defer replayTS.Close()
+
+	replayResp, err := http.Get(replayTS.URL + "/events")
+	if err != nil {
+		t.Fatalf("replay GET: %v", err)
+	}
+	replayBody, _ := io.ReadAll(replayResp.Body)
+	replayResp.Body.Close()
+
+	if replayResp.StatusCode != 200 {
+		t.Errorf("replay status = %d, want 200", replayResp.StatusCode)
+	}
+
+	// Check Content-Type.
+	ct := replayResp.Header.Get("Content-Type")
+	if ct != "text/event-stream" {
+		t.Errorf("Content-Type = %q, want text/event-stream", ct)
+	}
+
+	// The replayed body should contain SSE data lines.
+	if !strings.Contains(string(replayBody), "data:") {
+		t.Error("replayed response should contain SSE data")
+	}
+}
+
+// TestIntegration_SSE_Proxy_E2E tests the SSE proxy passthrough and L2
+// fallback flow end-to-end.
+func TestIntegration_SSE_Proxy_E2E(t *testing.T) {
+	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		flusher, ok := w.(http.Flusher)
+		if !ok {
+			http.Error(w, "no flusher", 500)
+			return
+		}
+		w.Header().Set("Content-Type", "text/event-stream")
+		w.WriteHeader(200)
+		flusher.Flush()
+
+		w.Write([]byte("data: proxy-event-1\n\n"))
+		flusher.Flush()
+		w.Write([]byte("data: proxy-event-2\n\n"))
+		flusher.Flush()
+	}))
+	defer upstream.Close()
+
+	l1 := NewMemoryStore()
+	l2 := NewMemoryStore()
+
+	proxy := NewProxy(l1, l2, WithProxyTransport(upstream.Client().Transport))
+
+	// Passthrough: forward to upstream.
+	req, _ := http.NewRequest("GET", upstream.URL+"/sse", nil)
+	resp, err := proxy.RoundTrip(req)
+	if err != nil {
+		t.Fatalf("RoundTrip: %v", err)
+	}
+	body, _ := io.ReadAll(resp.Body)
+	resp.Body.Close()
+
+	if !strings.Contains(string(body), "proxy-event-1") {
+		t.Error("proxy should deliver live SSE events")
+	}
+
+	// Wait for onDone save.
+	deadline := time.After(2 * time.Second)
+	for {
+		l2Tapes, _ := l2.List(t.Context(), Filter{})
+		if len(l2Tapes) > 0 {
+			break
+		}
+		select {
+		case <-deadline:
+			t.Fatal("L2 tape not saved within timeout")
+		default:
+			time.Sleep(10 * time.Millisecond)
+		}
+	}
+
+	// L2 fallback: upstream is now "down".
+	proxy2 := NewProxy(l1, l2,
+		WithProxyTransport(roundTripperFunc(func(_ *http.Request) (*http.Response, error) {
+			return nil, errors.New("down")
+		})),
+		WithProxySSETiming(SSETimingInstant()),
+	)
+
+	req2, _ := http.NewRequest("GET", upstream.URL+"/sse", nil)
+	resp2, err := proxy2.RoundTrip(req2)
+	if err != nil {
+		t.Fatalf("RoundTrip fallback: %v", err)
+	}
+	defer resp2.Body.Close()
+
+	src := resp2.Header.Get("X-Httptape-Source")
+	if src != "l1-cache" && src != "l2-cache" {
+		t.Errorf("X-Httptape-Source = %q, want l1-cache or l2-cache", src)
+	}
+
+	// Read and verify the fallback SSE body.
+	fallbackBody, _ := io.ReadAll(resp2.Body)
+	if !strings.Contains(string(fallbackBody), "proxy-event-1") {
+		t.Errorf("fallback should contain cached events, got: %q", string(fallbackBody))
 	}
 }
--- a/server.go
+++ b/server.go
@@ -327,5 +327,5 @@ func (s *Server) serveSSE(w http.ResponseWriter, r *http.Request, tape Tape) {
 	flusher.Flush()
 
 	// Replay events with the configured timing.
-	_ = replaySSEEvents(r.Context(), w, flusher, tape.Response.SSEEvents, s.sseTiming)
+	_ = replaySSEEvents(r.Context(), w, flusher, tape.Response.SSEEvents, s.sseTiming) //nolint:errcheck // SSE replay write failure is not actionable
 }
--- a/server.go
+++ b/server.go
@@ -16,14 +16,15 @@ import (
 type Server struct {
 	store          Store
 	matcher        Matcher
-	fallbackStatus int              // HTTP status when no tape matches
-	fallbackBody   []byte           // response body when no tape matches
+	fallbackStatus int                 // HTTP status when no tape matches
+	fallbackBody   []byte              // response body when no tape matches
 	onNoMatch      func(*http.Request) // optional callback when no tape matches
-	cors           bool             // if true, add CORS headers to all responses
-	delay          time.Duration    // fixed delay before every response; zero means no delay
-	errorRate      float64          // fraction of requests that return 500 (0.0-1.0)
-	randFloat      func() float64   // random number generator (injectable for testing)
-	replayHeaders  map[string]string // headers injected into every replayed response
+	cors           bool                // if true, add CORS headers to all responses
+	delay          time.Duration       // fixed delay before every response; zero means no delay
+	errorRate      float64             // fraction of requests that return 500 (0.0-1.0)
+	randFloat      func() float64      // random number generator (injectable for testing)
+	replayHeaders  map[string]string   // headers injected into every replayed response
+	sseTiming      SSETimingMode       // controls SSE replay inter-event timing
 }
 
 // ServerOption configures a Server.
@@ -109,6 +110,17 @@ func WithReplayHeaders(key, value string) ServerOption {
 	}
 }
 
+// WithSSETiming sets the inter-event timing mode for SSE tape replay.
+// Defaults to SSETimingRealtime (replay with original inter-event gaps).
+//
+// Modes:
+//   - SSETimingRealtime(): replay with original timing
+//   - SSETimingAccelerated(factor): divide gaps by factor (must be > 0)
+//   - SSETimingInstant(): emit all events immediately, no delay
+func WithSSETiming(mode SSETimingMode) ServerOption {
+	return func(s *Server) { s.sseTiming = mode }
+}
+
 // withRandFloat overrides the random number generator for testing.
 // This is unexported -- only used in tests to make error simulation
 // deterministic.
@@ -139,6 +151,7 @@ func NewServer(store Store, opts ...ServerOption) *Server {
 		matcher:        DefaultMatcher(),
 		fallbackStatus: http.StatusNotFound,
 		fallbackBody:   []byte("httptape: no matching tape found"),
+		sseTiming:      SSETimingRealtime(),
 	}
 	for _, opt := range opts {
 		opt(s)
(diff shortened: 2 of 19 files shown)


sse.go
```diff
@@ -0,0 +1,376 @@
+package httptape
```
tibtof, line 1: **Missing documentation (blocker)**: This PR does not include any documentation updates. Issue #124 has an explicit "Docs scope addendum" comment requiring:

- README updates (intro tagline, "Why httptape" bullet, "Use cases" subsection, "How it compares" table row, Quick start SSE replay snippet)
- `docs/recording.md`, `docs/replay.md`, `docs/sanitization.md`, `docs/proxy.md` updates
- `llms.txt` and `llms-full.txt` updates

SSE record/replay/sanitization is httptape's strongest differentiator vs every competitor (WireMock, gock, httpmock, MSW). Shipping the code without the documentation to surface that positioning defeats the marketing purpose of v0.10.0.

**Action**: Add all docs listed in the issue addendum. This is the top-priority fix.

server.go
```diff
@@ -258,20 +271,61 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	// Remove Content-Length — SSE streams are chunked.
+	w.Header().Del("Content-Length")
+
+	// Write status code.
+	w.WriteHeader(tape.Response.StatusCode)
+	flusher.Flush()
+
+	// Replay events with the configured timing.
+	_ = replaySSEEvents(r.Context(), w, flusher, tape.Response.SSEEvents, s.sseTiming)
```
tibtof: **Swallowed error without standard comment**: The error from `replaySSEEvents` is discarded with a bare `_ =`. Elsewhere in this file (lines 228, 295), the same pattern uses `//nolint:errcheck // response write failure is not actionable` to document why the error is intentionally dropped.

While discarding the error is defensible here (write failures in an HTTP handler are not actionable), it should follow the project's existing convention.

**Action**: Change to:
```go
_ = replaySSEEvents(r.Context(), w, flusher, tape.Response.SSEEvents, s.sseTiming) //nolint:errcheck // SSE replay write failure is not actionable
```

integration_test.go
```diff
@@ -643,3 +645,152 @@ func TestIntegration_RecordReplay_PostWithBody(t *testing.T) {
...
 }
+
+// TestIntegration_SSE_RecordReplay_E2E tests the full SSE record-and-replay
+// flow using MemoryStore: an upstream SSE server is recorded through the
+// Recorder, then replayed by the Server with instant timing.
+func TestIntegration_SSE_RecordReplay_E2E(t *testing.T) {
+	// Start an upstream that serves SSE.
+	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		flusher, ok := w.(http.Flusher)
+		if !ok {
```
tibtof, line 659: **Dead code and buggy upstream mock**: This test handler has two problems:

1. Dead code: `body := strings.NewReader(""); _ = body` serves no purpose.
2. Buggy data generation: `strings.TrimRight(strings.Repeat("0", i+1), "0")` always produces an empty string (trimming all `0` characters from a string composed entirely of `0` characters). Every event's data becomes `{"n":}` which is invalid JSON.

The test still passes because it only asserts `strings.Contains(body, "data:")`, which is too weak to catch this.

**Action**: Fix the upstream to produce valid SSE payloads (e.g., `fmt.Sprintf("data: {\"n\":%d}\n\n", i)`) and tighten the assertion to verify the replayed events contain meaningful data.

tibtof (commented):
## Automated Review: CHANGES_REQUESTED

### Blocking

**1. Missing documentation (all docs deliverables)**
Issue #124 has an explicit "Docs scope addendum" requiring README updates, `docs/recording.md`, `docs/replay.md`, `docs/sanitization.md`, `docs/proxy.md` updates, and `llms.txt`/`llms-full.txt` updates. Zero docs files are in the diff. SSE record/replay/sanitization is httptape's headline differentiator for v0.10.0. Shipping code without documentation that surfaces the feature's positioning defeats the release's purpose.

**2. Missing ADR-35 in `decisions.md`**
The architect's design was communicated via issue comments, and the issue status says "Design: ADR-35 in decisions.md", but `decisions.md` was not updated. The ADR should be added to maintain the project's decision log.

### Non-blocking (fix alongside docs)

**3. `server.go:330` -- Swallowed error without standard comment**
The `_ = replaySSEEvents(...)` discards the error without the project's standard `//nolint:errcheck` annotation. Other identical patterns in the same file document the intent.

**4. `integration_test.go` -- `TestIntegration_SSE_RecordReplay_E2E` has dead code and generates malformed JSON**
The upstream mock has unused variables and a `strings.TrimRight(strings.Repeat("0", ...), "0")` expression that always produces an empty string, making every event's data `{"n":}` (invalid JSON). The assertions are too weak to catch this.

### What looks good

The core implementation is solid:
- `SSEEvent`, `SSETimingMode` (sealed interface with 3 constructors), `sseRecordingReader` -- all per architect spec
- Critical `onDone` before `close(r.done)` ordering -- correct
- `health.go` properly refactored to use shared `writeSSEEvent` -- no duplicated SSE wire-format code
- Schema backward compatibility test present and passing
- 1752-line test suite covering parsing edge cases, timing modes, upstream disconnect, sanitization, proxy passthrough, proxy L2 fallback, concurrent recording/replay
- Race detector clean with `-count=5`
- 93.3% overall coverage; `sse.go` functional coverage is effectively 100% (uncovered lines are seal methods)
- No new dependencies (`go.mod`/`go.sum` unchanged)
- No panics in hot paths (only constructor guard in `SSETimingAccelerated`)
- `deepCopyTape` properly deep-copies `SSEEvents`
- All exported symbols have godoc comments
- `go vet` clean; `gofmt` issues are pre-existing

tibtof:
## Review threads addressed

### 1. Missing documentation (blocker -- resolved)

All documentation deliverables from the "Docs scope addendum" on issue #124 are now included:

**README.md updates:**
- Intro tagline updated to mention SSE streams
- New "Why httptape" bullet: SSE record/replay with LLM streaming as canonical use case
- New "Use cases" subsection: "Recording LLM streaming responses" with runnable code snippet
- New "SSE replay" quick start snippet using `WithSSETiming(SSETimingInstant())`
- New "SSE record/replay" row in the "How it compares" table (httptape `yes` vs WireMock `no`, gock `no`, MSW `partial`)

**docs/*.md updates:**
- `docs/recording.md` -- new "SSE (Server-Sent Events) recording" section covering detection, tape format, disabling, and redaction integration
- `docs/replay.md` -- new "SSE replay" section with timing modes table and LLM streaming test example
- `docs/sanitization.md` -- new "SSE event redaction" section with `RedactSSEEventData`, `FakeSSEEventData`, and complete LLM pipeline example
- `docs/proxy.md` -- new "SSE passthrough and fallback" section covering live passthrough, L2 fallback timing, and SSE redaction in proxy mode

**LLM-targeted summaries:**
- `llms.txt` -- updated overview, core types, recording, redaction, replay, proxy sections; added SSE comparison table; added SSE API functions
- `llms-full.txt` -- updated overview, key features, "Why httptape" section; added SSE sections within recording, replay, sanitization, and proxy pages; added SSE API functions to key APIs

### 2. Missing ADR-35 in `decisions.md` (blocker -- resolved)

Appended ADR-35 after ADR-29 (the latest existing ADR). Covers:
- Type design (SSEEvent, SSETimingMode sealed interface)
- Schema migration (RecordedResp.SSEEvents, mutual exclusion, backward compat)
- Recording approach (sseRecordingReader, io.TeeReader + io.Pipe, onDone ordering)
- Replay timing modes
- Sanitization composition (new SanitizeFunc constructors)
- Shared writeSSEEvent helper (health.go refactor)

### 3. `server.go:330` errcheck comment (non-blocking -- fixed)

Changed `_ = replaySSEEvents(...)` to include `//nolint:errcheck // SSE replay write failure is not actionable`, matching the convention at lines 228 and 295.

### 4. `integration_test.go` dead code and broken event data (non-blocking -- fixed)

- Removed dead code (`body := strings.NewReader(""); _ = body`)
- Fixed `strings.TrimRight(strings.Repeat("0", i+1), "0")` (always empty) to `fmt.Sprintf("data: {\"n\":%d}\n\n", i)` which produces valid JSON
- Tightened assertions: each event's payload is now verified individually (`data: {"n":0}` through `data: {"n":4}`) instead of just checking for `"data:"` substring

### Verification

- `go build ./...` -- clean
- `go test ./... -race -count=10` -- clean (23s)
- `go vet ./...` -- clean
- `gofmt -l server.go integration_test.go` -- clean
- `go.mod` / `go.sum` -- unchanged (no new dependencies)

tibtof (commented):
**APPROVED** (posted as comment because GitHub prevents self-approval)

Round 2 review -- all four items from round 1 are resolved:

1. **Documentation**: All deliverables landed -- README (intro, Why httptape bullet, LLM streaming use case, SSE replay quick start, comparison table row), docs/recording.md, docs/replay.md, docs/sanitization.md, docs/proxy.md, llms.txt, llms-full.txt. All code snippets reference real exported API names (`RedactSSEEventData`, `FakeSSEEventData`, `WithSSETiming`, `SSETimingInstant`, `WithSSERecording`, `WithProxySSETiming`) -- no fabricated symbols.

2. **ADR-35**: Appended to decisions.md. Covers all six required topics: SSEEvent type design, schema migration, sseRecordingReader recording approach, replay timing modes, sanitization composition, and writeSSEEvent helper.

3. **server.go errcheck nit**: Line 330 now has `//nolint:errcheck` matching the project convention.

4. **integration_test.go fixes**: Dead code removed; events now carry distinct valid JSON payloads (`{"n":0}` through `{"n":4}`); assertions verify each event individually.

Regression check: `go test ./... -race -count=10` clean (24s). SSE implementation files (sse.go, sse_test.go) untouched in round 2. No new dependencies (go.mod/go.sum unchanged).

Ready to merge -- no further rounds needed.

PR #166, opened 2026-04-18
feat(store): pluggable filename strategy for FileStore (readable filenames opt-in) (#132)

### LINKED ISSUE #132: feat: pluggable filename strategy for FileStore (readable filenames opt-in)
## Problem

`FileStore` currently names every tape `<uuid>.json` (`store_file.go:225`). Safe (no collisions ever) but opaque — humans browsing a fixtures directory or grep-ing a `.httptape-cache/` for a specific request have no readable cue. The `tape.ID` UUID is what every consumer uses internally, so the filename is essentially just a unique key on disk.

The `ts-frontend-first` example (#131) makes this visible: hand-curated upstream fixtures live as `get-products.json` / `get-profile.json` (a human picked the names), but the runtime L2 cache is a wall of UUIDs.

## Proposal

Add a pluggable filename strategy as a `FileStoreOption`. Two strategies built in:

```go
// Default — current behavior, no change for existing users.
WithFilenameStrategy(httptape.UUIDFilenames())

// Opt-in: readable filenames, collision-safe via short body-hash suffix.
WithFilenameStrategy(httptape.ReadableFilenames())
```

### `ReadableFilenames` format

```
<method>_<url-slug>[_<bodyhash-prefix>].json
```

| Request | Filename |
|---|---|
| `GET /api/products` | `get_api-products.json` |
| `GET /api/products/1` | `get_api-products-1.json` |
| `POST /api/cart` (body `{}`) | `post_api-cart_44ad.json` |
| `HEAD /` | `head_root.json` |
| `GET /api/products?limit=10` | `get_api-products_q-3a1f.json` (query short-hashed) |

Rules:
- Method lowercased.
- Path: leading `/` dropped, remaining `/` → `-`, percent-decoded characters whitelisted to `[a-z0-9-]`, anything else → `-`.
- Empty path becomes `root`.
- Body hash prefix: first 4 hex chars of `Tape.Request.BodyHash`. Appended only when body is non-empty (the well-known empty-body hash adds noise).
- Query string: short-hashed and prefixed `q-` (full querystrings are too long and often contain whitespace).
- Filename collision is theoretically possible but vanishingly unlikely with 4 hex of body hash. If a collision is detected on `Save`, append a `_2`, `_3`, … counter as a tie-breaker.

### Custom strategies

`FilenameStrategy` is a function type so users can plug their own:

```go
type FilenameStrategy func(tape Tape) string
```

A custom strategy that wants to fall back to UUID for collision safety can wrap one of the built-ins.

## Constraints

- **Backward compat is hard**: existing FileStore users have UUID-named files on disk. The opt-in flag must not silently rename them. Default behavior unchanged.
- **`Load(id)` and `Delete(id)` must still work** — both lookup by `tape.ID`. The strategy only affects the *write path*. On load, `FileStore` may need to scan-and-match, OR maintain an `id → filename` index sidecar file. Settle in design.
- **`List(filter)` must still return all tapes**, regardless of strategy used to write them. Mixed UUID + readable filenames in the same directory should be supported (a user might switch strategies mid-life, or hand-curate alongside generated fixtures).
- stdlib only.
- Race-clean tests, table-driven, ≥90% coverage on new code.

- **Lookup mechanism**: scan-and-match (read every file's `id` field on `Load`) is simple but O(n). An on-disk index (`.httptape-index.json` mapping `id → filename`) is fast but adds a moving piece. Alternative: encode the `id` into the filename as a suffix (`get_api-products_44ad__<short-id>.json`) so lookup is filesystem-glob-able. Pick one with reasoning.
- **Query string handling**: short-hash (proposed) vs full-encode vs strip-and-document. Short-hash loses information; full-encode produces unwieldy filenames. Could be controllable via strategy options.
- **Path slug aggressiveness**: keep `_` in slugs? `:` from path params? Trade readability vs filesystem safety.
- **Per-route subdirectories** (`<route>/<filename>`): the BACKLOG.md hint suggested this; it's not in scope here. Confirm explicit non-goal or fold in.

## Out of scope

- Per-route directory partitioning.
- Renaming existing on-disk fixtures (one-shot migration tool could be a follow-up).
- Configurable file extension (always `.json`).
- Filename strategies for memory store (irrelevant).

## Acceptance criteria

- New `FilenameStrategy` type and `WithFilenameStrategy` option exported.
- Built-in `UUIDFilenames()` (default) and `ReadableFilenames()` strategies.
- `Save`, `Load`, `List`, `Delete` all work correctly with both strategies.
- Mixed-strategy directories (some UUID-named, some readable) work for `Load` and `List`.
- Default behavior of `NewFileStore(WithDirectory(...))` is byte-for-byte unchanged.
- Tests cover collision handling, special characters in paths, query strings, empty paths, mixed-strategy directories.

tibtof:
Closes #132

## Summary

- Adds `FilenameStrategy` function type (`func(tape Tape) string`) and `WithFilenameStrategy` option for `FileStore`
- Two built-in strategies:
  - `UUIDFilenames()` -- default, byte-identical to existing `<id>.json` behavior
  - `ReadableFilenames()` -- `<method>_<url-slug>[_q-<query-hash>][_<body-hash>].json`
- `Load(id)` and `Delete(id)` use fast path (`<id>.json`) first, then scan directory on miss
- Filename collision resolution via `_2`, `_3`, ..., `_99` counter suffixes
- Save-overwrite cleanup: when re-saving a tape whose computed filename changed (e.g. strategy switch), the old file is removed
- Corrupt JSON files are silently skipped during scan (library convention: no logging)
- List now also skips corrupt JSON silently (was previously an error)

### Example: UUID vs Readable for the same tape

```
# UUID (default)
550e8400-e29b-41d4-a716-446655440000.json

# Readable
get_api-users.json
post_api-users_a1b2.json
get_api-users_q-3f4a.json
head_root.json
```

### Backward compatibility

- Default behavior is byte-identical to today's `<id>.json` filenames
- Mixed-strategy directories (some UUID files, some readable files) work seamlessly
- Existing fixtures require zero migration

## Test plan

- [x] `TestFilenameStrategy_UUIDDefault` -- default behavior produces `<id>.json` byte-identically
- [x] `TestFilenameStrategy_Readable_Examples` -- table-driven, covers GET, POST with body, GET with query, HEAD, DELETE, PUT with special chars
- [x] `TestFilenameStrategy_Readable_Slugify` -- edge cases: unicode, control chars, very long paths, empty path, all-special-chars
- [x] `TestFilenameStrategy_Readable_VeryLongPath` -- filename truncated to 200 chars
- [x] `TestFilenameStrategy_Readable_QueryHash` -- deterministic, different queries produce different filenames
- [x] `TestFilenameStrategy_Readable_BodyHash` -- different bodies produce different filenames
- [x] `TestFileStore_LoadByID_FastPath` -- UUID-named file loaded via O(1) fast path
- [x] `TestFileStore_LoadByID_ScanFallback` -- readable-named file found via directory scan
- [x] `TestFileStore_DeleteByID_BothPaths` -- fast path and scan fallback for Delete
- [x] `TestFileStore_MixedStrategyDirectory` -- CRUD works with mixed UUID + readable files
- [x] `TestFileStore_FilenameCollision` -- `_2` suffix added on collision
- [x] `TestFileStore_FilenameCollision_ExceedsLimit` -- error returned beyond `_99`
- [x] `TestFileStore_OverwriteWithChangedStrategy` -- old file removed on strategy switch
- [x] `TestFileStore_BackwardCompat` -- pre-populated UUID files work with new FileStore
- [x] `TestFileStore_CorruptJSONScanSkipped` -- corrupt JSON skipped during scan
- [x] `TestFileStore_EmptyStrategyFallback` -- empty strategy output falls back to tape ID
- [x] All existing tests pass unchanged
- [x] `go test -race -count=10` clean
- [x] `go vet` clean

### DIFF
--- a/store_file.go
+++ b/store_file.go
@@ -249,8 +249,13 @@ func (fs *FileStore) Save(ctx context.Context, tape Tape) error {
 
 	// Determine the target filename using the strategy.
 	stem := fs.strategy(tape)
-	if stem == "" {
-		// Defensive fallback: empty strategy output -> use tape ID.
+
+	// Sanitize strategy output to prevent path traversal from custom strategies.
+	// Built-in strategies (UUIDFilenames, ReadableFilenames) produce safe values,
+	// but user-provided strategies could return "../escape" or similar.
+	stem = filepath.Base(stem)
+	if stem == "" || stem == "." || stem == ".." {
+		// Defensive fallback: empty, ".", or ".." strategy output -> use tape ID.
 		stem = tape.ID
 	}
 
@@ -364,6 +369,8 @@ func (fs *FileStore) removeOldFile(tapeID, newFilename string) {
 			continue // corrupt JSON, skip
 		}
 		if id == tapeID {
+			// Best-effort: if removal fails (e.g., concurrent delete), the orphan
+			// file is left behind but Save still succeeds with the new file.
 			os.Remove(filepath.Join(fs.dir, name))
 			return // found and removed; IDs are unique
 		}
@@ -494,6 +501,10 @@ func (fs *FileStore) Delete(ctx context.Context, id string) error {
 	}
 
 	for _, entry := range entries {
+		if err := ctx.Err(); err != nil {
+			return fmt.Errorf("httptape: filestore delete %s: %w", id, err)
+		}
+
 		if entry.IsDir() || !strings.HasSuffix(entry.Name(), ".json") {
 			continue
 		}
@@ -562,9 +573,10 @@ func (fs *FileStore) scanForID(ctx context.Context, id string) (Tape, bool, erro
 	return Tape{}, false, nil
 }
 
-// extractIDFromJSON extracts the "id" field from a JSON byte slice without
-// fully unmarshalling the entire structure. This is used for efficient
-// scanning of directory contents.
+// extractIDFromJSON extracts the "id" field from a JSON byte slice by
+// unmarshalling into a minimal struct with only the ID field. The rest of
+// the document is parsed but discarded. This is used for scanning directory
+// contents when only the tape ID is needed.
 func extractIDFromJSON(data []byte) (string, error) {
 	var partial struct {
 		ID string `json:"id"`
--- a/store_file.go
+++ b/store_file.go
@@ -2,9 +2,12 @@ package httptape
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -15,6 +18,163 @@ import (
 // directory traversal components that could escape the base directory.
 var ErrInvalidID = errors.New("httptape: invalid tape ID")
 
+// ErrFilenameCollision is returned when a filename collision cannot be
+// resolved after exhausting the counter suffix range (_2 through _99).
+var ErrFilenameCollision = errors.New("httptape: filename collision limit exceeded")
+
+// maxCollisionCounter is the maximum counter suffix attempted when
+// resolving filename collisions.
+const maxCollisionCounter = 99
+
+// maxFilenameLength is the maximum number of characters allowed in a
+// filename before the ".json" extension. This keeps filenames under
+// filesystem limits (typically 255 bytes).
+const maxFilenameLength = 200
+
+// FilenameStrategy is a function that computes the filename stem (without
+// the ".json" extension) for a given Tape. The returned string must not
+// contain path separators. Built-in strategies are provided by
+// UUIDFilenames and ReadableFilenames.
+type FilenameStrategy func(tape Tape) string
+
+// UUIDFilenames returns a FilenameStrategy that uses the tape's ID as the
+// filename stem. This is the default strategy and produces filenames
+// identical to the original behavior (e.g., "550e8400-e29b-41d4-a716-446655440000.json").
+func UUIDFilenames() FilenameStrategy {
+	return func(tape Tape) string {
+		return tape.ID
+	}
+}
+
+// ReadableFilenames returns a FilenameStrategy that produces human-readable
+// filenames based on the tape's HTTP method, URL path, query string, and
+// request body hash.
+//
+// The format is: <method>_<url-slug>[_q-<query-hash>][_<body-hash>].json
+//
+// Examples:
+//
+//	GET /api/users              -> get_api-users.json
+//	POST /api/users (with body) -> post_api-users_a1b2.json
+//	GET /api/users?page=2       -> get_api-users_q-3f4a.json
+//	HEAD /                      -> head_root.json
+func ReadableFilenames() FilenameStrategy {
+	return func(tape Tape) string {
+		method := strings.ToLower(tape.Request.Method)
+		if method == "" {
+			method = "unknown"
+		}
+
+		slug := slugifyURL(tape.Request.URL)
+
+		parts := []string{method, slug}
+
+		// Append query hash if query string is non-empty.
+		if qh := queryHash(tape.Request.URL); qh != "" {
+			parts = append(parts, "q-"+qh)
+		}
+
+		// Append body hash if body is non-empty.
+		if bh := bodyHashPrefix(tape.Request.BodyHash); bh != "" {
+			parts = append(parts, bh)
+		}
+
+		name := strings.Join(parts, "_")
+
+		// Truncate to maxFilenameLength.
+		if len(name) > maxFilenameLength {
+			name = name[:maxFilenameLength]
+			// Clean up any trailing dash from truncation.
+			name = strings.TrimRight(name, "-")
+		}
+
+		if name == "" {
+			return tape.ID
+		}
+		return name
+	}
+}
+
+// slugifyURL extracts the path from a URL and normalizes it to a slug
+// containing only lowercase alphanumeric characters and dashes.
+// Returns "root" for empty or "/" paths.
+func slugifyURL(rawURL string) string {
+	path := ""
+	if u, err := url.Parse(rawURL); err == nil {
+		path = u.Path
+	} else {
+		// Fallback: try to extract path manually.
+		path = rawURL
+		if i := strings.Index(path, "://"); i >= 0 {
+			path = path[i+3:]
+		}
+		if i := strings.IndexByte(path, '/'); i >= 0 {
+			path = path[i:]
+		} else {
+			path = ""
+		}
+		if i := strings.IndexByte(path, '?'); i >= 0 {
+			path = path[:i]
+		}
+	}
+
+	// Strip leading and trailing slashes.
+	path = strings.Trim(path, "/")
+	if path == "" {
+		return "root"
+	}
+
+	// Normalize: replace any non-[a-z0-9] character with a dash,
+	// collapse consecutive dashes, trim leading/trailing dashes.
+	var b strings.Builder
+	b.Grow(len(path))
+	lastDash := false
+	for _, r := range strings.ToLower(path) {
+		if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') {
+			b.WriteRune(r)
+			lastDash = false
+		} else {
+			if !lastDash && b.Len() > 0 {
+				b.WriteByte('-')
+				lastDash = true
+			}
+		}
+	}
+
+	result := strings.TrimRight(b.String(), "-")
+	if result == "" {
+		return "root"
+	}
+	return result
+}
+
+// queryHash returns the first 4 hex characters of the SHA-256 hash of the
+// raw query string from the given URL. Returns "" if the query is empty.
+func queryHash(rawURL string) string {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return ""
+	}
+	raw := u.RawQuery
+	if raw == "" {
+		return ""
+	}
+	h := sha256.Sum256([]byte(raw))
+	return hex.EncodeToString(h[:2])
+}
+
+// bodyHashPrefix returns the first 4 hex characters of the given body hash.
+// Returns "" if the hash is empty.
+func bodyHashPrefix(fullHash string) string {
+	if fullHash == "" {
+		return ""
+	}
+	if len(fullHash) < 4 {
+		return fullHash
+	}
+	return fullHash[:4]
+}
+
 // FileStore is a filesystem-backed Store implementation. Each tape is persisted
 // as a single JSON file.
 //
@@ -22,8 +182,9 @@ var ErrInvalidID = errors.New("httptape: invalid tape ID")
 // process. It is not safe for multi-process concurrent access to the same
 // directory.
 type FileStore struct {
-	dir string // base directory for fixtures
-	mu  sync.RWMutex
+	dir      string           // base directory for fixtures
+	strategy FilenameStrategy // filename strategy for new saves
+	mu       sync.RWMutex
 }
 
 // FileStoreOption configures a FileStore.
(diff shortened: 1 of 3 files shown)


store_file.go
```diff
@@ -184,18 +473,106 @@ func (fs *FileStore) Delete(ctx context.Context, id string) error {
...
 
-	if err := os.Remove(path); err != nil {
+	// Scan fallback: look through all JSON files for a matching ID.
+	entries, err := os.ReadDir(fs.dir)
+	if err != nil {
 		return fmt.Errorf("httptape: filestore delete %s: %w", id, err)
 	}
-	return nil
+
+	for _, entry := range entries {
```
tibtof, line 503: **Missing context cancellation check**: Unlike `scanForID` (line 537) and `List` (line 436), the `Delete` scan fallback loop does not check `ctx.Err()` between iterations. If a directory contains many JSON files, this loop cannot be interrupted by a cancelled context.

**Why it matters**: This is inconsistent with the rest of the file. All other scan loops in this PR respect context cancellation. A directory with hundreds of tapes would scan all of them even after the caller has given up.

**Suggestion**: Add a `ctx.Err()` check at the top of the loop body, matching the pattern in `scanForID`:
```go
for _, entry := range entries {
    if err := ctx.Err(); err != nil {
        return fmt.Errorf("httptape: filestore delete %s: %w", id, err)
    }
    if entry.IsDir() || !strings.HasSuffix(entry.Name(), ".json") {
        continue
    }
    // ...
}
```

store_file.go
```diff
@@ -89,16 +279,101 @@ func (fs *FileStore) Save(ctx context.Context, tape Tape) error {
...
+		data, err := os.ReadFile(filepath.Join(fs.dir, name))
+		if err != nil {
+			continue
+		}
+		id, err := extractIDFromJSON(data)
+		if err != nil {
+			continue // corrupt JSON, skip
+		}
+		if id == tapeID {
+			os.Remove(filepath.Join(fs.dir, name))
```
tibtof, line 374: **Swallowed error on `os.Remove`**: The return value of `os.Remove` is discarded here. The project constitution states 'never swallow errors' (`fmt.Errorf("context: %w", err)`).

**Why it matters**: If the old file cannot be removed (e.g., permission denied), the caller has no way to know that stale data remains on disk. The `Save` succeeds, but now there are two files for the same tape ID, which will cause ambiguity on future `Load` or `List` calls.

**Suggestion**: Since `removeOldFile` is documented as best-effort, this is acceptable *if* we at least propagate the failure back to `Save` so the caller can decide. Consider either:
1. Returning an error from `removeOldFile` and handling it in `Save` (preferred), or
2. If truly best-effort, add a comment explicitly acknowledging the swallowed error with rationale (e.g., `// Best effort: failing to remove the old file is non-fatal because the new file will be written regardless.`)

store_file.go
```diff
@@ -72,6 +247,21 @@ func (fs *FileStore) Save(ctx context.Context, tape Tape) error {
 	fs.mu.Lock()
 	defer fs.mu.Unlock()
 
+	// Determine the target filename using the strategy.
+	stem := fs.strategy(tape)
```
tibtof, line 251: **No path-traversal validation on strategy output**: The stem returned by `fs.strategy(tape)` is used directly in `resolveFilename` which builds a file path via `filepath.Join(fs.dir, candidate)`. If a user provides a custom `FilenameStrategy` that returns a string containing `../` or path separators, the resulting file could be written outside the base directory.

**Why it matters**: The `validateID` function carefully guards against path traversal in tape IDs, but the strategy output (which becomes the actual filename) bypasses this check entirely. While the built-in strategies (`UUIDFilenames`, `ReadableFilenames`) are safe, the `FilenameStrategy` type is exported and users are encouraged to provide custom implementations. This is a path-traversal vulnerability.

**Suggestion**: After computing the stem, validate it the same way `validateID` validates IDs. At minimum, reject stems containing `/`, `\`, or `..` components:
```go
stem := fs.strategy(tape)
if stem == "" {
    stem = tape.ID
}
// Sanitize strategy output to prevent path traversal.
stem = filepath.Base(stem) // strip any directory components
if stem == "." || stem == ".." {
    stem = tape.ID
}
```
Alternatively, call `validateID(stem)` and fall back to `tape.ID` if it fails.

store_file.go
```diff
@@ -184,18 +473,106 @@ func (fs *FileStore) Delete(ctx context.Context, id string) error {
...
+		return tape, true, nil
+	}
+
+	return Tape{}, false, nil
+}
+
+// extractIDFromJSON extracts the "id" field from a JSON byte slice without
+// fully unmarshalling the entire structure. This is used for efficient
+// scanning of directory contents.
+func extractIDFromJSON(data []byte) (string, error) {
```
tibtof, line 580: **Misleading godoc -- `json.Unmarshal` does parse the full document**: The comment says 'without fully unmarshalling the entire structure', but `json.Unmarshal` tokenizes the entire JSON input even when the target struct only has one field. For tape files with large bodies (e.g., multi-MB response bodies), this means the scan path reads and fully parses every JSON document in the directory just to extract a small string.

**Why it matters**: During `Load` with scan fallback or `removeOldFile`, each file in the directory undergoes a full JSON parse. This is functionally correct, but the godoc creates a false impression of efficiency that may discourage future optimization.

**Suggestion** (non-blocking -- just fix the godoc): Either:
1. Update the comment to say 'by unmarshalling into a minimal struct (only the id field is extracted)', which is accurate, or
2. Replace with `json.NewDecoder` token-by-token parsing that truly stops after finding the `id` key, for a genuine performance improvement on large files.

tibtof (commented):
## Review: CHANGES REQUESTED -- PR #166 (feat: pluggable FileStore filename strategy)

### What was verified and passes

**Backward compatibility** -- CONFIRMED:
- `UUIDFilenames()` is the default in `NewFileStore` (line 213)
- `UUIDFilenames` produces `<tape.ID>.json` -- byte-identical to the old `tapePath` behavior
- `TestFilenameStrategy_UUIDDefault` and `TestFileStore_BackwardCompat` are both present and meaningful -- they pre-populate UUID-named files and verify Load/List/Save/Delete all work with a new `FileStore` instance
- Old fixtures continue to load without any code changes from users

**Tests** -- STRONG:
- Coverage at 93.5% (above 90% threshold)
- `go test -race -count=10` passes cleanly
- Table-driven tests used throughout (slugify, bodyHashPrefix, extractIDFromJSON)
- No test dependencies, stdlib `testing` only
- Comprehensive coverage: fast-path, scan-fallback, collision resolution, strategy switching, overwrite cleanup, corrupt JSON handling, mixed-strategy directories

**Code quality** -- MOSTLY GOOD:
- Every exported symbol has godoc
- `go vet` clean
- No `panic` anywhere
- No new dependencies (stdlib only)
- No global variables or `init()` side effects
- Functional options pattern for `WithFilenameStrategy`

**Strategy correctness** -- CONFIRMED:
- `ReadableFilenames` format matches spec: `<method>_<url-slug>[_q-<query-hash>][_<body-hash>].json`
- `slugifyURL` handles edge cases correctly (empty path, unicode, control chars, consecutive dashes)
- Query hash is 4 hex chars of SHA-256, deterministic
- Body hash prefix is first 4 chars of `tape.Request.BodyHash`
- Filename capped at 200 chars before `.json`
- Collision resolution: `_2` through `_99`, then `ErrFilenameCollision`

**List behavior change** (corrupt JSON silently skipped):
- The change is reasonable -- a single corrupt file should not break `List()`
- Note: ADR-36 is referenced in the issue but does not exist in `decisions.md`. The authorization for this behavior change should be documented.

### Blocking issues (2)

1. **Security: No path-traversal validation on strategy output** (line 251 in `store_file.go`). A custom `FilenameStrategy` returning `"../escape"` would write files outside the base directory. The `validateID` function guards tape IDs but the strategy output bypasses it. Apply `filepath.Base()` or equivalent sanitization.

2. **Consistency: `Delete` scan loop missing `ctx.Err()` check** (line 496 in `store_file.go`). Unlike `scanForID` and `List`, the `Delete` scan fallback does not check for context cancellation between iterations.

### Non-blocking issues (2)

3. **Swallowed `os.Remove` error** in `removeOldFile` (line 367). Either propagate the error or add an explicit comment documenting the best-effort semantics.

4. **Misleading godoc** on `extractIDFromJSON` (line 565-567). Says "without fully unmarshalling" but `json.Unmarshal` parses the full document. Fix the comment or switch to `json.NewDecoder` for genuine partial parsing.

tibtof:
## Review fixes addressed

All four review items from CHANGES_REQUESTED have been addressed in commit bb18b7b:

### Blocking issues

1. **Path-traversal vulnerability in strategy output** (line 251) -- FIXED. Strategy output is now sanitized with `filepath.Base()` to strip directory traversal components. After sanitization, empty / `"."` / `".."` results fall back to `tape.ID`. Added two test functions:
   - `TestFilenameStrategy_PathTraversalSanitized` -- verifies malicious strategies (`../escape`, `../../etc/passwd`, `/etc/passwd`, `..\..\escape`, `sub/dir`) all write inside the base directory.
   - `TestFilenameStrategy_EmptyOrDotResultFallsBack` -- verifies `""`, `"."`, `".."` all fall back to `tape.ID + ".json"`.

2. **Delete scan-loop missing context cancellation check** (line 496) -- FIXED. Added `ctx.Err()` check at the top of the loop body, matching the pattern in `scanForID` and `List`. Added test `TestFileStore_Delete_ScanFallback_ContextCancelled` confirming `Delete(ctx, id)` returns `context.Canceled` when ctx is cancelled.

### Non-blocking issues

3. **Swallowed `os.Remove` error in `removeOldFile`** (line 367) -- Added comment documenting the best-effort semantics: `// Best-effort: if removal fails (e.g., concurrent delete), the orphan file is left behind but Save still succeeds with the new file.`

4. **Misleading godoc on `extractIDFromJSON`** (lines 565-567) -- Fixed comment to accurately describe the behavior: `// extractIDFromJSON extracts the "id" field from a JSON byte slice by unmarshalling into a minimal struct with only the ID field. The rest of the document is parsed but discarded.`

### ADR-36

Appended ADR-36 to `decisions.md` documenting the pluggable FileStore filename strategy design. Covers all resolved questions: lookup mechanism, query handling, slug aggressiveness, body-hash length, collision counter, filename length cap, save overwrite cleanup, empty-strategy fallback, corrupt-JSON skip, and path-traversal sanitization.

### Verification

- `go build ./...` -- clean
- `go vet ./...` -- clean
- `gofmt` -- clean
- `go test ./... -race -count=10` -- all pass

tibtof:
## Review: APPROVED -- PR #166 (round 2)

Round 2 verification complete -- all four issues from round 1 are properly fixed. Ready to merge, no further rounds needed.

### Blocker fixes verified

1. **Path-traversal sanitization** (line 256 of `store_file.go`): `filepath.Base(stem)` strips directory components from strategy output before joining with the base directory. Empty/`.`/`..` post-sanitization results fall back to `tape.ID`. Test `TestFilenameStrategy_PathTraversalSanitized` covers 5 attack vectors (`../escape`, `../../etc/passwd`, `/etc/passwd`, `..\..\escape`, `sub/dir`) -- each confirms the file lands inside the base directory and does not escape. `TestFilenameStrategy_EmptyOrDotResultFallsBack` covers 3 edge cases.

2. **Delete ctx.Err() check** (line 504 of `store_file.go`): Placed at the top of the scan loop body, before any file I/O, matching the pattern in `scanForID` and `List`. Test `TestFileStore_Delete_ScanFallback_ContextCancelled` asserts specifically for `context.Canceled` and confirms the tape was not deleted.

### Nit fixes verified

3. `removeOldFile` (lines 372-373): Clear best-effort comment documents the intentional fire-and-forget semantics.

4. `extractIDFromJSON` (lines 576-579): Godoc now accurately describes that the document is unmarshalled into a minimal struct with only the ID field.

### ADR-36

Verified in `decisions.md`. Covers all required topics: lookup mechanism, query string handling, slug aggressiveness, body hash prefix length, collision counter, filename length cap, save overwrite cleanup, empty/dot fallback, corrupt-JSON skip, and path-traversal sanitization.

### Regression check

- `go test ./... -race -count=10` -- all pass
- `go vet ./...` -- clean
- No dependency changes (`go.mod`/`go.sum` untouched)
- Only 3 files modified: `decisions.md`, `store_file.go`, `store_file_test.go`

PR #182, opened 2026-04-18
fix(matcher): MatchBodyFuzzy vacuous-true for body-less requests (#178)

### LINKED ISSUE #178: fix: MatchBodyFuzzy eliminates body-less requests (both-empty should be vacuously true)
## Problem statement

`MatchBodyFuzzy` returns score 0 when neither the incoming request nor the candidate tape has a body (e.g., a `GET /users/1` matched against a recorded `GET /users/1`). This eliminates the candidate, making `MatchBodyFuzzy` unsafe to compose globally in a `CompositeMatcher` alongside REST-style body-less routes.

## User story

As a Go developer using `CompositeMatcher` with `MatchBodyFuzzy`, I want body-less requests to pass through the fuzzy body criterion without being eliminated, so that I can safely compose `MatchBodyFuzzy` alongside method/path criteria for a mix of `GET` and `POST` routes.

## Root cause

In `matcher.go` lines 421-477, `MatchBodyFuzzy` returns 0 when `matched == 0` (no fields compared). This is correct when there **should** have been a body to compare but wrong when **neither side has a body at all** — the match is vacuously true.

Specifically:
- When `reqBody` is nil/empty, `json.Unmarshal` fails and returns 0.
- Even if both are empty, there are no fields to compare, so `matched == 0 → return 0`.

## Acceptance criteria

1. When both the incoming request body AND the candidate tape's recorded body are empty (nil, zero-length, or not valid JSON), `MatchBodyFuzzy` returns a small positive score (1) — enough to keep the candidate alive without dominating other criteria.
2. When exactly one side has a body and the other does not, `MatchBodyFuzzy` returns 0 (mismatch, unchanged behavior).
3. When both sides have valid JSON bodies and the specified fields match, `MatchBodyFuzzy` returns 6 (unchanged behavior).
4. When both sides have valid JSON bodies and the specified fields differ, `MatchBodyFuzzy` returns 0 (unchanged behavior).
5. New table-driven tests added in `matcher_test.go` covering all four cases above.
6. Existing tests in `matcher_test.go` continue to pass unmodified.
7. Godoc comment on `MatchBodyFuzzy` updated to document the vacuous-true case for body-less requests.

## Out of scope

- Changes to any other criterion (`MatchBodyHash`, `MatchMethod`, etc.).
- Changes to `CompositeMatcher` scoring logic.
- Any new criterion types.

## Open questions

None.

tibtof:
Closes #178

## Summary

- `MatchBodyFuzzy` now returns score 1 (vacuous match) when both the incoming request body and the tape body are absent (nil, empty, or not valid JSON), instead of returning 0 which incorrectly eliminated body-less requests (GET, DELETE, HEAD) from matching.
- When exactly one side is absent, the criterion returns 0 (asymmetric mismatch).
- When both sides have valid JSON, the existing field-extraction and `reflect.DeepEqual` logic is unchanged (returns 6 on match, 0 on mismatch).
- Implements ADR-37 from `decisions.md`.

## Test plan

- [x] Updated `TestMatchBodyFuzzy_BothBodiesEmpty` expected score from 0 to 1
- [x] Added `TestMatchBodyFuzzy_VacuousTrue` table-driven test with 13 sub-cases:
  - both nil -> 1
  - both empty bytes -> 1
  - req nil / tape empty -> 1
  - req empty / tape nil -> 1
  - both invalid JSON -> 1
  - req nil / tape has body -> 0
  - req has body / tape nil -> 0
  - req invalid / tape has body -> 0
  - req has body / tape invalid -> 0
  - both empty JSON objects `{}` -> 0 (valid JSON, no paths match)
  - both JSON `null` -> 0 (valid JSON, paths fail extraction)
  - both bodied fields match -> 6 (unchanged)
  - both bodied fields differ -> 0 (unchanged)
- [x] All existing `TestMatchBodyFuzzy_*` tests pass unchanged
- [x] `go test ./...` passes
- [x] `go vet ./...` passes

### DIFF
--- a/matcher.go
+++ b/matcher.go
@@ -389,8 +389,9 @@ func headerContains(h http.Header, canonicalKey, value string) bool {
 //     (nil, empty, or not valid JSON), the criterion returns 1 (vacuous
 //     match — the body dimension is irrelevant for this request/tape
 //     pair). If exactly one side is absent, the criterion returns 0.
-//   - Both bodies are unmarshaled as JSON. If either body is not valid JSON,
-//     the criterion returns 0 (no match).
+//   - When both bodies are present (not absent per the rule above),
+//     they are unmarshaled as JSON. Path extraction and comparison
+//     proceeds on the unmarshaled values.
 //   - For each specified path, the value is extracted from both the request
 //     and the tape body. If a path does not exist in both bodies, it is
 //     skipped (does not cause a mismatch).
--- a/matcher.go
+++ b/matcher.go
@@ -385,6 +385,10 @@ func headerContains(h http.Header, canonicalKey, value string) bool {
 //   - $.array[*].field    -- field within each element of an array
 //
 // Matching semantics:
+//   - If both the incoming request body and the tape body are absent
+//     (nil, empty, or not valid JSON), the criterion returns 1 (vacuous
+//     match — the body dimension is irrelevant for this request/tape
+//     pair). If exactly one side is absent, the criterion returns 0.
 //   - Both bodies are unmarshaled as JSON. If either body is not valid JSON,
 //     the criterion returns 0 (no match).
 //   - For each specified path, the value is extracted from both the request
@@ -404,7 +408,8 @@ func headerContains(h http.Header, canonicalKey, value string) bool {
 //
 // Invalid or unsupported paths are silently ignored (same as RedactBodyPaths).
 //
-// Returns score 6 on match, 0 on mismatch.
+// Returns score 6 on match, 1 on vacuous match (both bodies absent),
+// 0 on mismatch.
 //
 // Note: using both MatchBodyFuzzy and MatchBodyHash in the same
 // CompositeMatcher is safe but semantically redundant. If MatchBodyHash
@@ -443,12 +448,17 @@ func MatchBodyFuzzy(paths ...string) MatchCriterion {
 			reqBody = bodyBytes
 		}
 
-		// Unmarshal both bodies.
+		// Determine whether each side is "absent" (nil, empty, or not valid JSON).
 		var reqData, tapeData any
-		if err := json.Unmarshal(reqBody, &reqData); err != nil {
-			return 0
+		reqAbsent := len(reqBody) == 0 || json.Unmarshal(reqBody, &reqData) != nil
+		tapeAbsent := len(candidate.Request.Body) == 0 || json.Unmarshal(candidate.Request.Body, &tapeData) != nil
+
+		// Vacuous-true: both bodies absent — return minimum positive score.
+		if reqAbsent && tapeAbsent {
+			return 1
 		}
-		if err := json.Unmarshal(candidate.Request.Body, &tapeData); err != nil {
+		// Asymmetric: one absent, one present — mismatch.
+		if reqAbsent || tapeAbsent {
 			return 0
 		}
 
--- a/matcher_test.go
+++ b/matcher_test.go
@@ -1286,6 +1286,48 @@ func TestMatchBodyFuzzy_VacuousTrue(t *testing.T) {
 	}
 }
 
+func TestMatchBodyFuzzy_VacuousTrueComposability(t *testing.T) {
+	// Integration test: validates the real user story from issue #178.
+	// A CompositeMatcher with MatchMethod, MatchPath, and MatchBodyFuzzy
+	// must correctly match a body-less GET request against a body-less
+	// GET tape, without MatchBodyFuzzy eliminating the candidate via score 0.
+	matcher := NewCompositeMatcher(
+		MatchMethod(),
+		MatchPath(),
+		MatchBodyFuzzy("$.action"),
+	)
+
+	// Candidate tapes: one body-less GET and one bodied POST.
+	getTape := Tape{
+		ID: "get-tape",
+		Request: RecordedReq{
+			Method: "GET",
+			URL:    "http://example.com/test",
+			Body:   nil, // no body
+		},
+	}
+	postTape := Tape{
+		ID: "post-tape",
+		Request: RecordedReq{
+			Method: "POST",
+			URL:    "http://example.com/test",
+			Body:   []byte(`{"action":"create"}`),
+		},
+	}
+	candidates := []Tape{getTape, postTape}
+
+	// Incoming request: GET /test with no body.
+	req := httptest.NewRequest("GET", "/test", nil)
+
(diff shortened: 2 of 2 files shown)


matcher.go
```diff
@@ -385,6 +385,10 @@ func headerContains(h http.Header, canonicalKey, value string) bool {
 //   - $.array[*].field    -- field within each element of an array
 //
 // Matching semantics:
+//   - If both the incoming request body and the tape body are absent
+//     (nil, empty, or not valid JSON), the criterion returns 1 (vacuous
+//     match — the body dimension is irrelevant for this request/tape
+//     pair). If exactly one side is absent, the criterion returns 0.
 //   - Both bodies are unmarshaled as JSON. If either body is not valid JSON,
 //     the criterion returns 0 (no match).
```
tibtof: **Contradictory godoc**: This line now contradicts the new vacuous-true semantics added at lines 388-391. When *both* bodies are not valid JSON, the criterion returns 1 (vacuous match), not 0. But this line still says "If either body is not valid JSON, the criterion returns 0 (no match)."

This will confuse readers because the two adjacent bullet points give opposite answers for the same scenario (both bodies are invalid JSON).

**Suggestion**: Reword this to clarify it applies only when one side has been parsed successfully, e.g.:

```go
//   - When both bodies are present (not absent per the rule above),
//     they are unmarshaled as JSON. Path extraction and comparison
//     proceeds on the unmarshaled values.
```

Or simply delete the sentence since the new bullet above already covers the absent/not-valid-JSON case completely, and the remaining bullets describe the path-extraction logic that only runs when both sides have valid JSON.

matcher_test.go
```diff
@@ -1175,8 +1175,114 @@ func TestMatchBodyFuzzy_BothBodiesEmpty(t *testing.T) {
...
+			req := httptest.NewRequest("POST", "/test", body)
+			tape := Tape{Request: RecordedReq{Body: tt.tapeBody}}
+
+			got := criterion(req, tape)
+			if got != tt.want {
+				t.Errorf("MatchBodyFuzzy() = %d, want %d", got, tt.want)
+			}
+		})
 	}
 }
```
tibtof, line 1288: **Missing ADR-37-specified test**: ADR-37 explicitly requires a `TestMatchBodyFuzzy_VacuousTrueComposability` integration test that validates the real user story from issue #178. The ADR specifies:

> Add `TestMatchBodyFuzzy_VacuousTrueComposability` that creates a `CompositeMatcher` with `MatchMethod()`, `MatchPath()`, and `MatchBodyFuzzy("$.action")`, then verifies that a `GET /users` request (no body) correctly matches a recorded `GET /users` tape (no body) while a `POST /users` tape with a body is not incorrectly selected.

This test is important because it validates the fix at the integration level -- ensuring the vacuous-true score actually solves the reported problem (body-less requests being eliminated from a `CompositeMatcher`). The existing unit tests validate the criterion in isolation, but do not prove the fix works end-to-end through the matcher.

**Suggestion**: Add a test here that creates a `CompositeMatcher`, populates it with a body-less `GET /users` tape and a bodied `POST /users` tape, then asserts that a body-less `GET /users` request matches the correct tape.

tibtof (commented):
## Review: Changes Requested

Two issues found -- one documentation correctness problem and one missing test:

1. **Contradictory godoc (matcher.go line 392-393)**: The pre-existing bullet "Both bodies are unmarshaled as JSON. If either body is not valid JSON, the criterion returns 0" now contradicts the newly added vacuous-true bullet directly above it. When both bodies are not valid JSON, the criterion returns 1 (per the new logic), not 0 (per the old text). This needs to be reworded or removed to avoid misleading readers.

2. **Missing `TestMatchBodyFuzzy_VacuousTrueComposability` test**: ADR-37 explicitly specifies this integration-level test to validate the real user story (a `CompositeMatcher` with `MatchMethod()`, `MatchPath()`, and `MatchBodyFuzzy("$.action")` correctly matching a body-less `GET /users` against a recorded body-less `GET /users` tape). The existing unit tests validate the criterion in isolation but do not prove the fix works through the `CompositeMatcher`.

Everything else looks correct: the implementation logic matches ADR-37, the vacuous-true score is 1, the "absent" definition correctly uses `len(body) == 0 || json.Unmarshal fails`, valid JSON values (`{}`, `null`, `[]`) correctly fall through to normal path-extraction, the existing `TestMatchBodyFuzzy_BothBodiesEmpty` test is updated to expect 1, the table-driven test covers all 13 edge cases from the ADR, no new dependencies, all tests pass, and only the two expected files are modified.

matcher.go
tibtof: Fixed in c8c6633. Replaced the contradictory bullet with:

```
//   - When both bodies are present (not absent per the rule above),
//     they are unmarshaled as JSON. Path extraction and comparison
//     proceeds on the unmarshaled values.
```

The old wording ("If either body is not valid JSON, the criterion returns 0") is now gone. The godoc reads top-to-bottom consistently: absent/absent -> 1, one absent -> 0, both present -> unmarshal + path-extraction.

matcher_test.go
tibtof: Added TestMatchBodyFuzzy_VacuousTrueComposability in c8c6633. The test creates a CompositeMatcher with MatchMethod() + MatchPath() + MatchBodyFuzzy("$.action"), provides two candidate tapes (a body-less GET /test and a bodied POST /test), and asserts that a body-less GET /test request matches the GET tape. Without the vacuous-true fix, MatchBodyFuzzy would return 0 for the GET tape, causing CompositeMatcher to eliminate it.

tibtof:
## Re-review: APPROVED

Re-reviewed commit c8c6633 against both previously requested changes:

1. **Godoc fix verified**: The stale bullet ("If either body is not valid JSON, returns 0") is fully replaced with accurate three-part wording covering vacuous-true (both absent -> 1), asymmetric mismatch (one absent -> 0), and normal comparison (both present -> unmarshal and extract). The summary line at the bottom also correctly lists all three return values. No contradictions remain.

2. **TestMatchBodyFuzzy_VacuousTrueComposability verified**: The test exists with the exact required name, exercises a CompositeMatcher with MatchMethod + MatchPath + MatchBodyFuzzy("$.action"), provides two candidates (body-less GET, bodied POST), sends a body-less GET request, and asserts the GET tape is correctly returned. All 13 sub-cases in TestMatchBodyFuzzy_VacuousTrue plus the composability test pass.

Full test suite passes. No new issues found. Ready for human review.

PR #184, opened 2026-04-18
feat: declarative matcher composition via config file (#180)

tibtof:
Closes #180

## Summary

Implements ADR-39: declarative matcher composition via JSON config file. This enables container consumers (Java/Kotlin/TS demos) to configure the httptape replay server's matching strategy without code changes, by extending the existing `Config` struct with an optional `matcher` field.

### Key design decisions (per ADR-39)

1. **Single `CriterionConfig` struct with type discriminator** -- mirrors the existing `Rule` pattern. `Type` field dispatches to the right criterion constructor. Only `Paths` is needed as a type-specific field (for `body_fuzzy`).

2. **`criterionBuilders` dispatch table** -- maps criterion type names to validate/build function pairs. Extensible for future criterion types without changing the dispatch logic.

3. **Scope limited to `method`, `path`, `body_fuzzy`** -- other types (`path_regex`, `route`, `headers`, `query_params`, `body_hash`) are follow-up issues per architect decision.

4. **`rules` field relaxed** -- `"rules": []` is now valid when a matcher is configured. Existing sanitization-only configs continue to work unchanged.

### Changes

- **`config.go`**: Added `MatcherConfig`, `CriterionConfig` types. Added `Matcher *MatcherConfig` field to `Config`. Added `BuildMatcher()` method with `criterionBuilders` dispatch table. Extended `Validate()` with matcher validation. Relaxed empty-rules check.
- **`config.schema.json`**: Added `matcher` property with `criteria` array schema. Added `if/then` constraints for per-type field validation. Removed `minItems: 1` from `rules`.
- **`cmd/httptape/main.go`**: Wired `--config` flag in `runServe` to load config and build matcher via `WithMatcher()`. Updated flag description. Removed dead TODO comment.
- **`config_test.go`**: 12 new tests covering matcher config loading, validation errors, `BuildMatcher` default/composed/error cases, backward compatibility.
- **`cmd/httptape/main_test.go`**: 3 new tests covering serve with valid matcher config (integration), invalid config, and no config.
- **`integration_test.go`**: End-to-end test with two `POST /v1/chat/completions` fixtures distinguished by `$.messages[*].role` via body_fuzzy criterion loaded from config.
- **`decisions.md`**: ADR-39 added.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `go test -race ./...` passes
- [x] Config with no matcher field returns `DefaultMatcher()` (backward compat)
[...]

Issue #180, opened 2026-04-18
feat: declarative matcher composition via config file (and wire --config into serve)


tibtof:
> **Note:** This issue replaces the previous `--match-body-fuzzy` CLI flag design. The flag-based approach was rejected because it does not compose well and leads to flag explosion as more criteria are exposed. Instead, we extend the existing JSON config infrastructure (`Config` in `config.go`, validated via `config.schema.json`) to declare matcher composition, and we wire the already-present but unused `--config` flag in `cmd/httptape/main.go` `runServe` into the server.

## Problem statement

Container consumers (Java/Kotlin/TS demos) cannot today configure the httptape server's matcher beyond the hardcoded `DefaultMatcher()` (method + path). The Kotlin/Koog agent demo needs body-aware matching to distinguish multiple LLM calls at the same URL. Adding per-criterion CLI flags (`--match-body-fuzzy`, future `--match-headers`, etc.) does not compose cleanly. The existing `Config` JSON infrastructure already supports declarative configuration for sanitization rules and should be extended to also declare matcher composition.

## User story

As a Go developer running `httptape serve` in a Docker container, I want to provide a JSON config file that declares which matching criteria the server uses, so that I can compose method + path + body-fuzzy (or other criteria in the future) without learning new CLI flags for each one.

## Motivation

The upcoming Kotlin + Ktor + Koog + Kotest + Testcontainers demo features a multi-step AI agent:
1. User asks a question, agent emits a `tool_call` (first `POST /v1/chat/completions`, body has `[system, user]` messages).
2. App executes the tool locally, agent does a second LLM call with the tool result (second `POST /v1/chat/completions`, body has `[system, user, assistant(tool_call), tool]` messages).

Both requests share the same method, path, and headers. Only the `messages` array differs. A config file declaring `body_fuzzy` matching with `paths: ["$.messages[*].role"]` would produce different extracted arrays, allowing `CompositeMatcher` to select the correct fixture.

## Proposed config shape

Extend `Config` (in `config.go`) with an optional top-level `matcher` field:

```json
{
  "version": "1",
  "matcher": {
    "criteria": [
      {"type": "method"},
      {"type": "path"},
      {"type": "body_fuzzy", "paths": ["$.messages[*].role"]}
    ]
  },
  "rules": [ ... existing sanitization rules ... ]
}
```

[...]

Issue #179, opened 2026-04-18
refactor: promote MatchCriterion from function type to Criterion interface


tibtof:
## Problem statement

`MatchCriterion` is currently a function type (`type MatchCriterion func(req *http.Request, candidate Tape) int`). The top-level `Matcher` is already an interface (port), but criteria — the composable units that `CompositeMatcher` aggregates — are bare functions. This inconsistency prevents clean config-driven matcher composition, makes debugging opaque (functions have no identity), and diverges from the project's hexagonal/interface-driven design principles.

## User story

As a Go developer building custom matchers, I want criteria to be expressed as an interface (`Criterion`) so that I can inspect, compose, and configure them in a type-safe, interface-driven way consistent with the rest of the httptape API.

## Proposed shape (for architect's consideration)

```go
type Criterion interface {
    Score(req *http.Request, candidate Tape) int
}

type CriterionFunc func(req *http.Request, candidate Tape) int
func (f CriterionFunc) Score(req *http.Request, candidate Tape) int { return f(req, candidate) }
```

Each existing criterion becomes a struct implementing `Criterion`:

| Current function | Proposed struct |
|---|---|
| `MatchMethod()` | `MethodCriterion{}` |
| `MatchPath()` | `PathCriterion{}` |
| `MatchPathRegex(pattern)` | `PathRegexCriterion{re}` |
| `MatchRoute(route)` | `RouteCriterion{Route}` |
| `MatchQueryParams()` | `QueryParamsCriterion{}` |
| `MatchHeaders(key, value)` | `HeadersCriterion{Key, Value}` |
| `MatchBodyHash()` | `BodyHashCriterion{}` |
| `MatchBodyFuzzy(paths...)` | `BodyFuzzyCriterion{Paths, parsed}` |

## Acceptance criteria

1. A `Criterion` interface is defined in `matcher.go` with a single method `Score(req *http.Request, candidate Tape) int`.
2. A `CriterionFunc` adapter type is defined, allowing ad-hoc functions to satisfy `Criterion`.
3. All 8 existing criteria are converted to exported struct types implementing `Criterion`.
4. `CompositeMatcher` accepts `[]Criterion` (via constructor) instead of `[]MatchCriterion`.
5. `DefaultMatcher()` updated to use the new struct types.
6. All existing tests in `matcher_test.go` updated and passing with no behavioral changes.
7. Godoc comments updated on all affected types and functions.
[...]

PR #186, opened 2026-04-18
feat(#185): Kotlin + Ktor + Koog + Kotest demo with body_fuzzy matching

### LINKED ISSUE #185: feat(examples): Kotlin + Ktor + Koog + Kotest demo with mocked OpenAI agent and weather REST
## Goal

Add `examples/kotlin-ktor-koog/` — a Kotlin + Ktor + Koog + Kotest application that demonstrates how to **test a single-tool AI agent and a classic REST integration deterministically** using httptape served via Testcontainers, with predefined fixtures.

This demo is the proof artifact that the three matcher improvements — #178 (vacuous-true fix), #179 (Criterion interface), and #180 (declarative matcher composition via config) — all work correctly together end-to-end.

## Why this demo matters

The matcher improvements (#178, #179, #180) were built specifically to unblock this scenario: a multi-step agent flow where **two POST requests hit the same URL** (`/v1/chat/completions`) and must be distinguished by JSON body content, while a **body-less REST GET** coexists in the same composite matcher. This is the first demo that exercises all three improvements together in a realistic scenario.

## Headline scenario: single-tool Koog agent + mocked weather REST

User asks: *"What's the weather in Berlin? Should I bring an umbrella?"*

### Wire-level trace (all three HTTP calls hit the **same** httptape container)

```
User → Ktor  GET /weather-advice?city=Berlin

  HTTP #1 → httptape (mocked OpenAI):  POST /v1/chat/completions
                                        messages=[system, user]
  HTTP #1 ← assistant(tool_call: getWeather("Berlin"))         [SSE]

  ─── Koog agent runs the getWeather tool ───
  HTTP A → httptape (mocked weather):   GET /v1/forecast?city=Berlin  (or wttr.in shape)
  HTTP A ← {"city":"Berlin","temp":12,"condition":"rain",...}  [JSON]

  HTTP #2 → httptape (mocked OpenAI):  POST /v1/chat/completions
                                        messages=[system, user, assistant(tc), tool(weather JSON)]
  HTTP #2 ← streaming "Yes, bring an umbrella..."             [SSE]

User ← Ktor SSE stream
```

**Key matching challenge solved by #178/#179/#180:**
- HTTP #1 and HTTP #2 share the same method (`POST`) and path (`/v1/chat/completions`). They are distinguished via `BodyFuzzyCriterion` on `$.messages[*].role` — HTTP #1 has only `[system, user]` roles; HTTP #2 has `[system, user, assistant, tool]`.
- HTTP A is a `GET` with no request body. It coexists in the same `CompositeMatcher` thanks to the #178 vacuous-true fix (both-empty body = score 1, not eliminated).
- The matcher composition is declared in `httptape.config.json` using the #180 config format.

### Three fixtures

| Fixture file | HTTP call | Method | Path | Response format |
|---|---|---|---|---|
| `chat-1.json` | #1 | POST | `/v1/chat/completions` | SSE with `tool_call` delta (getWeather) |
| `weather-berlin.json` | A | GET | `/v1/forecast?city=Berlin` (or wttr.in shape) | JSON |
| `chat-2.json` | #2 | POST | `/v1/chat/completions` | SSE streaming "Yes, bring an umbrella..." |

## Tech stack

**Hard constraint for architect/dev:** verify every dependency version against Maven Central / Gradle Plugin Portal at implementation time. Do not assume older versions. The Spring Boot 3.5 vs 4.0.5 incident in the Java demo is the cautionary tale — always check what "latest stable" actually is at the moment of implementation.

| Component | Target | Notes |
|---|---|---|
| Kotlin | 2.x latest stable | Verify against Gradle Plugin Portal |
| JDK | 25 | Matches the Java demo |
| Ktor | 3.x latest stable | Netty engine, coroutines |
| Koog | Latest stable | Was 0.5.1 at last check; **verify current** |
| Kotest | Latest stable | BehaviorSpec recommended (see open questions) |
| Testcontainers | Latest stable | Java API, called from Kotlin |
| Gradle | Latest stable | Kotlin DSL (`build.gradle.kts`), wrapper committed |
| httptape image | `ghcr.io/vibewarden/httptape:latest` or pinned | Architect to decide pinning strategy |

## Weather API choice

Use **wttr.in** (`https://wttr.in/Berlin?format=j1`) as the real-world weather API shape. Free, no API key, accepts city names directly, real-world wire format. Architect can substitute a fictional service if there is a strong reason (e.g., licensing concerns), but default to wttr.in.

## Acceptance criteria

- [ ] 1. **Directory**: `examples/kotlin-ktor-koog/` exists with a complete, buildable project.
- [ ] 2. **Build system**: Gradle Kotlin DSL (`build.gradle.kts`), wrapper committed (`gradlew`, `gradlew.bat`, `gradle/wrapper/`). Latest stable Gradle version.
- [ ] 3. **Ktor endpoint**: one HTTP endpoint `GET /weather-advice?city={city}` that returns the agent's streamed advice as SSE (Server-Sent Events).
- [ ] 4. **Koog agent**: agent definition with one tool `getWeather(city: String)` that calls a weather REST API.
- [ ] 5. **Configurable base URLs**: the tool implementation hits a weather URL whose base is configurable via environment variable or property, so tests can redirect traffic to httptape. Same for the OpenAI base URL.
- [ ] 6. **Fixtures**: three pre-recorded fixture files in `src/test/resources/fixtures/` (or per Kotest convention):
  - `chat-1.json` — OpenAI tool-call request, SSE response with `tool_call` delta for `getWeather`
  - `weather-berlin.json` — weather REST GET, JSON response
  - `chat-2.json` — OpenAI follow-up with tool result in messages, SSE streaming natural-language response
- [ ] 7. **Matcher config**: `httptape.config.json` declaring the composite matcher (`method` + `path` + `body_fuzzy` on `$.messages[*].role`), mounted into the container. This exercises the entire #178/#179/#180 stack end-to-end.
- [ ] 8. **Kotest test**: BehaviorSpec (or architect-chosen alternative) test: given a Koog agent backed by httptape, when the user asks the weather-advice question, then the response streams the expected text. The test orchestrates Testcontainers, the Ktor app, and the agent flow.
- [ ] 9. **Single httptape container**: one shared Testcontainers instance serves both the mocked OpenAI route and the mocked weather route (per-JVM lifecycle, like the Java demo's `TestcontainersConfig`).
- [ ] 10. **`api.http` file**: HTTP request file matching the Java demo's pattern, with the demo's HTTP requests.
- [ ] 11. **Standalone path**: `Dockerfile` + `docker-compose.yml` so that `docker compose up -d` followed by `curl` works without running tests.
- [ ] 12. **README**: follows the Java demo's structure:
  - What it shows
  - Headline scenario
  - Quick start (`./gradlew test`)
  - Dev workflow
  - Testcontainers reuse opt-in
  - "Try it standalone" (`docker compose up`)
  - "Stack" table
  - "Why not...?" comparison section
- [ ] 13. **README matcher callout**: README explicitly calls out that the demo exercises the matcher composition introduced in #178/#179/#180 — distinguishing two same-URL OpenAI calls plus a body-less REST GET via one composite matcher declared in config.
- [ ] 14. **Dependabot**: add a `gradle` ecosystem block to `.github/dependabot.yml` scoped to `examples/kotlin-ktor-koog/`, plus a `docker` block if there is a Dockerfile. Follow the existing pattern (weekly Monday 09:00 UTC, labels, minor-and-patch grouping).
- [ ] 15. **CI**: extend `.github/workflows/examples.yml` to include a `kotlin-ktor-koog` matrix entry with `setup-java` (JDK 25), Gradle cache, and `./gradlew test` as the build command.
- [ ] 16. **SSE fixture accuracy**: the two OpenAI SSE fixtures (`chat-1.json`, `chat-2.json`) must use OpenAI's exact `chat.completion.chunk` wire format, including `tool_calls` deltas in `chat-1.json`'s stream, so that Koog's OpenAI client deserializer does not choke.
- [ ] 17. **`./gradlew test` passes** from a fresh clone with JDK 25 + Docker running.
- [ ] 18. **PR workflow**: PR opened with full review cycle, **not merged** — held for the user's personal review.

## Out of scope

- **Real OpenAI or wttr.in calls in any code path** — recording is documented in the README but not automated in this demo.
- **Multi-tool agents** — single tool (`getWeather`) is sufficient to show the pattern.
- **Multi-turn conversation memory** — single turn is enough.
- **RAG, vector stores, embeddings** — separate concern.
- **Anthropic, Bedrock, Ollama, Gemini providers** — OpenAI wire format only.
- **Spring Boot, Micronaut, or other JVM frameworks** — this demo is pure Ktor + Koog.
- **httptape recording mode** — the demo uses pre-recorded fixtures in `serve` mode only.
- **CLI flag-based matcher composition** — the demo uses the config-file approach from #180; per-criterion CLI flags were explicitly rejected in the #180 design pivot.
- **Merging the PR** — the PR is held for the user's personal review.

1. **Test runner ergonomics**: Kotest BehaviorSpec vs FunSpec? PM recommendation is BehaviorSpec because the Given/When/Then structure maps naturally to the agent flow narrative ("Given a Koog agent backed by httptape / When the user asks the weather-advice question / Then the response streams the expected text"). Architect can override with rationale.

2. **Koog tool implementation idiom**: Koog has its own Tool DSL. Confirm at architect time whether the tool's HTTP client should use Ktor client, OkHttp, or whatever Koog itself uses internally for HTTP calls. The choice affects how the base URL is injected.

3. **Multi-module vs single-module Gradle layout**: single module is simpler and matches the demo's scope; multi-module might be overkill. Architect's call.

4. **Matcher config file location and mounting**: where does `httptape.config.json` live in the source tree? Likely `src/test/resources/httptape.config.json`. How is it mounted into the container — `withCopyFileToContainer`? What path does the `--config` flag reference inside the container? Architect to specify the exact mount path and the CLI flag value.

5. **SSE fixture authoring strategy**: the two OpenAI fixtures need real `chat.completion.chunk` JSON in OpenAI's exact wire format, including the `tool_calls` deltas in `chat-1.json`'s stream. Options: (a) capture from real OpenAI, (b) hand-craft following the OpenAI API reference, (c) borrow from the Java demo's existing `chat-completion-headphones.json` fixture as a starting template and adapt for tool-call format. Architect should specify the approach.

## References

- **Matcher improvements this demo exercises**: #178 (vacuous-true), #179 (Criterion interface), #180 (declarative config)
- **Java demo for parallel structure**: #167
- **Java demo directory**: `examples/java-spring-boot/`
- **Dependabot pattern**: `.github/dependabot.yml` (existing Maven + Docker blocks for Java demo)
- **CI pattern**: `.github/workflows/examples.yml` (existing Java + TS matrix entries)

## Status

`READY_FOR_ARCH` — PM spec complete.

tibtof:
Closes #185

## Summary

- Add complete `examples/kotlin-ktor-koog/` demo (22 files) that exercises httptape's declarative matcher composition (#178, #179, #180) in a realistic Koog AI agent scenario
- A Ktor server exposes `GET /weather-advice?city={city}` backed by a Koog `singleRunStrategy` agent with a weather tool
- Three httptape fixtures (2x POST /v1/chat/completions + 1x GET /v1/forecast) are served by a single container, with `body_fuzzy` on `$.messages[*].role` distinguishing the two POST requests
- Testcontainers builds the httptape image from the repo root Dockerfile via `ImageFromDockerfile` so the test always uses the `--config` serve feature (not yet in a published release)
- Add Gradle + Docker blocks to `.github/dependabot.yml` and a matrix entry to `.github/workflows/examples.yml` with Gradle cache support
- Update `examples/README.md` with the new example row

## Test plan

- [x] `./gradlew test --no-daemon` passes locally (Kotest BehaviorSpec asserts the agent response contains "umbrella")
- [x] `go build ./...`, `go test ./...`, `go vet ./...` all pass (Go library unchanged)
- [ ] CI workflow `build (kotlin-ktor-koog)` passes in GitHub Actions
- [ ] Verify `docker compose up -d --build && curl -N http://localhost:8080/weather-advice?city=Berlin && docker compose down` works

## Key design decisions

- **Build from source**: The `--config` flag in `serve` mode was merged to main (#184) but is not in any published Docker image yet. `ImageFromDockerfile` builds httptape from the repo root Dockerfile so the test is self-contained and always exercises the latest matcher code.
- **Non-streaming fixtures**: Koog's `singleRunStrategy` uses `stream=false` in its OpenAI requests, so fixtures use standard `chat.completion` JSON format (not SSE chunks).
- **Gradle cache in CI**: The `setup-java` action's cache parameter is conditionally set to `gradle` or `maven` based on the matrix entry's `gradle-cache` flag.


examples/kotlin-ktor-koog/README.md
```diff
@@ -0,0 +1,110 @@
+# Test your Koog AI agents deterministically
+
+Working example of [httptape](https://github.com/VibeWarden/httptape) used to test a **Koog single-tool AI agent** with a weather REST integration -- both served from pre-recorded fixtures via [Testcontainers](https://testcontainers.com/), with zero real API calls.
+
+## What this demo shows
+
+A Ktor server hosts a `GET /weather-advice?city={city}` endpoint that runs a Koog AI agent. The agent calls OpenAI's chat completions API (non-streaming) and a weather REST API, then returns weather-based advice. Tests exercise the full agent loop against a single httptape container serving hand-crafted fixtures.
+
```
tibtof, line 8: **Missing build-from-source documentation**: The README says "(builds httptape from source)" in the "Try it standalone" section, but does not explain **why** building from source is necessary. Users who clone this demo will not understand that `--config` support (PR #184) is not yet in any published release (v0.10.1 predates it).

Add a note in the README (e.g., under "Prerequisites" or as a callout near "Quick start") that explicitly states:

> **Note:** This demo requires httptape with `--config` support in `serve` mode, which landed in PR #184. Until v0.11.0 is released, the demo builds httptape from source via Testcontainers (`ImageFromDockerfile`) and Docker Compose (`build: context: ../..`). Once v0.11.0 is published, these can be switched back to `ghcr.io/vibewarden/httptape:0.11.0`.

This is the difference between a demo that "just works" for users and one that silently confuses them when they try to use a published image instead.

examples/kotlin-ktor-koog/src/main/resources/application.yaml
```diff
@@ -0,0 +1,11 @@
...
+  deployment:
+    port: 8080
+
+koog:
+  openai:
+    base-url: ${OPENAI_BASE_URL:-https://api.openai.com}
+    api-key: ${OPENAI_API_KEY:-sk-placeholder}
+
+weather:
+  base-url: ${WEATHER_BASE_URL:-https://wttr.in}
```
tibtof: **Dead configuration**: The `application.yaml` defines `koog.openai.base-url`, `koog.openai.api-key`, and `weather.base-url` config entries, but these are never read by the application code. `Application.kt` uses `System.getenv()` directly with default fallbacks, bypassing Ktor config file resolution entirely.

This is confusing for readers: the YAML config suggests the app reads these values from the config file, but it does not. Either:
1. Remove the `koog:` and `weather:` sections from `application.yaml` (since they are unused), or
2. Wire the application to actually read from `application.yaml` via `environment.config.property(...)` as the ADR specified.

Option 1 is simpler and avoids misleading readers. The `ktor.deployment.port` entry is still used by Ktor and should be kept.

examples/kotlin-ktor-koog/src/main/kotlin/dev/httptape/demo/WeatherTools.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Koog tool set providing a weather forecast lookup tool.
+ *
+ * The [baseUrl] is injected via constructor so tests can redirect
+ * HTTP calls to the httptape container.
+ */
+class WeatherTools(private val baseUrl: String) : ToolSet {
+
+    private val client = HttpClient(CIO) {
```
tibtof: **Resource leak**: The `HttpClient(CIO)` created at line 23 is never closed. Each call to `weatherAdviceRoute` creates a new `WeatherTools` instance (line 14 of `WeatherAdviceRoute.kt`), and each `WeatherTools` instance creates a new `HttpClient` that is never closed.

This means every request to `/weather-advice` leaks a CIO engine with its coroutine context. For a demo this may be tolerable, but it is bad practice to show in example code that users will copy.

Fix: Either (a) close the client after use via `client.use { ... }`, (b) make `WeatherTools` implement `Closeable` and close it after the agent run, or (c) make the `HttpClient` a parameter injected from a longer-lived scope (e.g., created once in `configureApp` and passed through).

tibtof:
## Automated Review -- CHANGES_REQUESTED

**Verdict: CHANGES_REQUESTED** (1 blocking issue, 2 non-blocking suggestions)

### Blocking

**README does not document the build-from-source requirement.** The `HttptapeContainer.kt` correctly uses `ImageFromDockerfile` to build httptape from the repo root Dockerfile, and the `docker-compose.yml` correctly uses `build: context: ../..`. The KDoc in `HttptapeContainer.kt` explains the rationale well. However, the README does not surface this information to users. The "Try it standalone" section says "(builds httptape from source)" parenthetically but never explains that `--config` support is not in any published release (v0.10.1 predates PR #184). Users who try to swap in `ghcr.io/vibewarden/httptape:0.10.1` will get a broken demo with no explanation. Add a visible note explaining the version gap and build-from-source requirement.

### Non-blocking suggestions

1. **Dead config in `application.yaml`**: The `koog:` and `weather:` sections are never read by the application code (which uses `System.getenv()` directly). Either remove them or wire the app to read from the config file. Leaving them is confusing for readers.

2. **`HttpClient` leak in `WeatherTools.kt`**: A new `HttpClient(CIO)` is created per request and never closed. For a demo, inject the client from a longer-lived scope or close it after use.

### What was reviewed and looks good

- **ADR-40 compliance**: All dependency versions match exactly (Kotlin 2.3.20, Ktor 3.4.2, Koog 0.8.0, Kotest 6.1.11, Testcontainers 2.0.4, Gradle 9.4.1, JDK 25). Single module, BehaviorSpec, Ktor CIO client for the weather tool, matcher config at the specified paths -- all per spec.
- **Build-from-source workaround implementation**: `ImageFromDockerfile` in Testcontainers and `build: context: ../..` in docker-compose.yml are both correct and functionally equivalent to what a production deployment with a published image would look like. The workaround does not affect demo correctness.
- **Fixture correctness**: chat-1.json (roles: [system, user]), chat-2.json (roles: [system, user, assistant, tool]), and weather-berlin.json (body-less GET) correctly exercise the body_fuzzy discrimination and vacuous-true matching. All base64 bodies decode to valid OpenAI wire-format JSON. Using non-streaming (`chat.completion`) instead of SSE (`chat.completion.chunk`) is a valid simplification allowed by the ADR.
- **Matcher config**: Correctly declares method + path + body_fuzzy on `$.messages[*].role`.
- **Kotlin idioms**: Idiomatic Kotlin throughout. Proper scope functions (`apply`, `also`), `object` singleton for container lifecycle, BehaviorSpec Given/When/Then reads cleanly, no nullable abuse, no `runBlocking` in production paths.
- **Test structure**: BehaviorSpec with `testApplication` is clean and matches the preferred approach from the ADR.
- **CI/dependabot**: Gradle and Docker ecosystem entries added correctly. The `gradle-cache` matrix property with conditional cache type selection is a clean solution.
- **File layout**: Matches ADR-40 except for the explicitly-optional files (AppConfig.kt, WeatherAgent.kt -- both folded into other files as the ADR permitted).
- **No Go library changes**: Correctly scoped to the examples directory plus CI/dependabot config.

---

3 inline comments posted. See individual file comments for details and suggested fixes.

examples/kotlin-ktor-koog/README.md
tibtof: Fixed in 4581df8. Added a dedicated "Note on httptape version" section after the matcher composition callout, explaining why building from source is required (--config from PR #184 not in any release), how the build-from-source works in both Testcontainers and docker-compose, and exactly what to change after v0.11.0 ships. Also updated the Quick start section and Stack table with first-run build time (~30s) and a row for httptape linking to the version note.

examples/kotlin-ktor-koog/src/main/resources/application.yaml
tibtof: Fixed in 4581df8. Removed the dead `koog:` and `weather:` sections from `application.yaml`. Only the `ktor.deployment.port` entry remains, which is the only one actually consumed by Ktor. The app reads all other config via `System.getenv()` with default fallbacks in `Application.configureApp()`.

examples/kotlin-ktor-koog/src/main/kotlin/dev/httptape/demo/WeatherTools.kt
tibtof: Fixed in 4581df8. Went with Option A (app-scoped client). The `HttpClient(CIO)` is now created once in `Application.configureApp()`, passed through `weatherAdviceRoute()` into `WeatherTools` as a constructor parameter, and closed via `monitor.subscribe(ApplicationStopped)`. No more per-request client creation or leak.

tibtof:
## Re-review: LGTM / Approved

All three items from the previous review are correctly addressed in commit 4581df8.

### 1. README version-gap documentation (was BLOCKING) -- Fixed

The "Note on httptape version" section is well-placed between the matcher composition callout and Prerequisites. It covers all required points:
- **Why build-from-source**: explains that `--config` support (PR #184) is not in any published release (v0.10.1 predates it).
- **How it works today**: documents `ImageFromDockerfile` in `HttptapeContainer.kt` and `build: context: ../..` in `docker-compose.yml`.
- **What changes after v0.11.0**: clear one-line swap instructions for both files.
- **Quick start** mentions ~30s first-run build time.
- **Stack table** has `httptape | built from source` row with a link back to the note.

### 2. Dead config entries in application.yaml (was non-blocking) -- Fixed

`application.yaml` now contains only `ktor.deployment.port: 8080`. The removed `koog:` and `weather:` sections are replaced by `System.getenv()` calls with sensible defaults in `configureApp()`. Clean and correct.

### 3. HttpClient resource leak in WeatherTools.kt (was non-blocking) -- Fixed

Properly implemented as app-scoped client (Option A):
- `HttpClient(CIO)` created once in `Application.configureApp()`.
- Closed on shutdown via `monitor.subscribe(ApplicationStopped) { weatherHttpClient.close() }`.
- Passed through `weatherAdviceRoute()` into `WeatherTools` constructor.
- `WeatherTools` no longer creates its own client -- it receives `client: HttpClient` via constructor injection.

No new issues found. Ready for human review.

tibtof:
Both review items addressed in commit 79142fa:

1. **FreeSpec**: `WeatherAdviceTest` now extends `FreeSpec` instead of `BehaviorSpec`. Uses the `- {}` hierarchy you prefer. Import updated to `io.kotest.core.spec.style.FreeSpec`. All assertion logic and Testcontainers wiring preserved exactly — only the spec style changed. `./gradlew test` passes.

2. **TestApplication runner**: Added `TestApplication.kt` in test sources with a `main()` function that boots `HttptapeContainer.instance`, reads `HttptapeContainer.baseUrl`, and starts the Ktor app on port 8080 with the same Testcontainers wiring the integration tests use. Added a `testRun` Gradle task (`JavaExec` on the test classpath, mainClass `dev.httptape.demo.TestApplicationKt`). README updated with a "Development workflow — run with the same Testcontainers setup" section. IntelliJ right-click on `TestApplication.kt` -> Run also works out of the box.

PR is NOT merged — held for your personal review.

tibtof:
Both points addressed in commit b401526:

1. **Flattened hierarchy**: Removed the four-level FreeSpec nesting. The test is now a single flat string -- no intermediate `-` parents. When more tests arrive, we can reintroduce grouping where it actually helps.

2. **Business-flavored name**: Renamed from `"streams the expected umbrella advice"` to `"advises bringing an umbrella when it is rainy in the requested city"`. Reads like a feature spec, not a test assertion. No implementation vocabulary (streams, expected, endpoint).

All assertion logic, fixture wiring, and Testcontainers setup are unchanged. `./gradlew test` passes.

PR #228, opened 2026-04-26
chore(#218): defensively validate tar entry names

### LINKED ISSUE #218: security: defensive tar entry validation in bundle.go isFixtureEntry
## Source

Discovered during the 2026-04-25 sanitization boundary audit.

## Finding

`isFixtureEntry` (bundle.go:290-293) checks tar entry prefix (`fixtures/`) and suffix (`.json`) only. It does not reject tar header names containing path-traversal patterns (e.g., `fixtures/../../../etc/passwd.json`).

## Risk assessment

**Not currently exploitable.** `ImportBundle` does not extract entries to the filesystem — it deserializes JSON into `Tape` objects in memory, then persists via `Store.Save` (which has its own path validation in FileStore).

**Why fix it anyway:** sanitization-boundary code deserves defensive validation. If a future change adds filesystem extraction (or a custom `Store` implementation that interprets `tape.ID` as a path without sanitizing), the missing check could become exploitable. Defense in depth.

## Acceptance criteria

- `isFixtureEntry` rejects names containing `..`, leading `/`, backslashes, or absolute paths.
- A test asserts that a malformed tar bundle (path-traversal entry) is rejected with a clear error.
- The check is documented as a defensive guard, not a security boundary in itself.

## References

- `bundle.go:290-293` (`isFixtureEntry`)
- Audit report: `~/notes/httptape/audit-2026-04-25-sanitization.md` (concern: Low / tar entry validation)

tibtof:
Closes #218

## Summary

- Add `validateEntryName` in `bundle.go` that rejects tar entry names containing path-traversal patterns (`..` as a path component), absolute paths (leading `/`), or backslashes (Windows-style separators).
- Call `validateEntryName` in `ImportBundle` before processing each tar entry. Malicious entries cause an immediate error return (fail loudly rather than silently skip).
- This is **defense in depth**, not a security patch. `ImportBundle` deserializes entries into in-memory `Tape` objects (not to the filesystem), and `Store.Save` implementations have their own path validation via `validateID`. The new check hardens the trust boundary so that no future code path could be exploitable.
- Uses `path.Clean` (not `filepath.Clean`) for OS-independent tar path normalization, then checks each segment for `..`.

## Test plan

- `TestImportBundle_RejectsPathTraversalInTarEntryNames` -- table-driven test with 6 cases covering `..` traversal, absolute paths, and backslashes; each constructs a tar bundle with the malicious entry and asserts `ImportBundle` returns the expected error.
- `TestImportBundle_AllowsSafeEntryNames` -- verifies that legitimate names (including `tape..1.json` where `..` is a substring, not a path component) are accepted.
- `TestValidateEntryName` -- unit tests for the validation function itself with 10 cases.

```
go test ./... -count=1   # all pass
go vet ./...             # clean
```

### DIFF
--- a/bundle.go
+++ b/bundle.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"path"
 	"sort"
 	"strings"
 	"time"
@@ -242,6 +243,10 @@ func ImportBundle(ctx context.Context, s Store, r io.Reader) error {
 			return fmt.Errorf("httptape: import: %w", err)
 		}
 
+		if err := validateEntryName(hdr.Name); err != nil {
+			return err
+		}
+
 		lr := io.LimitReader(tr, maxEntrySize)
 
 		switch {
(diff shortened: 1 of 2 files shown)


tibtof (commented):
**Reviewer verdict: APPROVE** (submitted as comment because GitHub prevents self-approval)

Reviewed against the full checklist:

**Validation logic** is correct: rejects absolute paths, backslashes, and `..` as a path component while correctly allowing `..` as a substring in filenames (e.g., `tape..1.json`). Uses `path.Clean` (not `filepath.Clean`), which is right because tar paths are always forward-slash regardless of OS.

**Hook point** is correct: validation runs in `ImportBundle` immediately after reading each tar header, before any entry processing (LimitReader, JSON decode).

**Error behavior** is correct: returns an error (fail loudly) rather than silently skipping malicious entries. Appropriate for a security boundary.

**Framing** is accurate: both the PR description and code comment explicitly state this is defense in depth, not a security patch, since ImportBundle deserializes into memory (not filesystem). No overclaiming.

**Tests** are thorough: 3 test functions covering 19 total cases -- unit-level validation function tests, integration-level rejection tests through ImportBundle, and positive tests confirming safe names are accepted. All table-driven, stdlib testing only.

**Scope** is minimal: only `bundle.go` and `bundle_test.go` modified, 177 lines added. No new dependencies (only `path` from stdlib). No architecture violations.

`go build`, `go test`, and `go vet` all pass clean.

PR #235, opened 2026-04-26
feat(#196): response template helpers with faker integration

### LINKED ISSUE #196: feat(templating): response template helpers with deterministic faker integration
## Problem statement

httptape already has `{{request.*}}` templating in `server.go` for response bodies and headers. Realistic test fixtures need more: timestamps, UUIDs, counters, random strings, and — crucially — deterministic fake data that stays consistent across related requests in a test. Additionally, fixtures that serve parameterized routes (e.g., `/users/:id`) need to extract path parameters and use them in response templates.

## User story

As a Go developer authoring fixtures, I want template helpers for timestamps, UUIDs, counters, deterministic fake data, and path-parameter accessors so that my mock responses contain realistic, dynamic values — including values derived from the request URL — without hand-coding them into every fixture.

## Context

The deterministic faker integration is the differentiator here. Other approaches offer non-deterministic random values; httptape can offer reproducible fakes via its existing HMAC-based Faker infrastructure (same seed produces the same output across runs). For multi-turn agent testing, this means a fake user ID generated in response 1 is the same fake user ID in response 3 — consistency without manual fixture wiring.

Path-parameter templating extends this further: a single tape for `/users/:id` can produce deterministic, distinct fake data for `/users/1`, `/users/2`, `/users/42` — each with its own stable fake name, email, etc., derived from the captured `:id` segment.

## Design

Extend the existing `{{...}}` parser with a curated set of helpers. Stdlib-only per project constraint.

### New helpers

```handlebars
{{now}}                          — RFC3339 timestamp at serve time
{{now format=unix}}              — unix timestamp
{{now format=iso}}               — shorthand alias for RFC3339
{{uuid}}                         — random UUID v4
{{randomHex length=16}}          — random hex string
{{randomInt min=0 max=100}}      — random integer
{{counter name=req-id}}          — per-name incrementing counter (server-scoped)
{{faker.email seed=user-42}}     — deterministic fake email
{{faker.name seed=user-42}}      — deterministic fake full name
{{faker.phone seed=user-42}}     — deterministic fake phone
{{faker.uuid seed=user-42}}      — deterministic fake UUID
{{faker.address seed=user-42}}   — deterministic fake address
```

### Key design point

`{{faker.*}}` reuses the existing HMAC-based Faker infrastructure. Same `seed` produces the same output across runs. This is the novel differentiator — reproducible fakes baked into templating.

The `{{faker.*}}` helpers should bridge to all existing Faker types documented in `docs/api-reference.md` (currently 12 built-in types).

### Counter behavior

Counter is per-server-instance, per-name, survives across requests within the server's lifetime. Resettable via admin endpoint (tie to request journal feature) or `Server.ResetCounter(name)`.

### Path-parameter accessors

New accessor `{{pathParam.NAME}}` resolves to the captured value for the named segment `NAME` from the URL pattern match. Example: for pattern `/users/:id/orders/:orderId`, a request to `/users/42/orders/7` makes `{{pathParam.id}}` resolve to `42` and `{{pathParam.orderId}}` resolve to `7`.

Existing `{{request.query.NAME}}` already works for query parameters — verify and document as a first-class accessor. `{{request.path}}` already works for the raw path.

### Nested template evaluation

The value of a `seed` argument (e.g., `seed=user-{{pathParam.id}}`) must be evaluated before the helper runs. This enables dynamic seed derivation from captured path parameters, which is the key mechanism for per-ID deterministic faking.

### Matcher side

New criterion `PathPatternCriterion` that matches URL patterns with named segments and exposes the capture map to the template evaluation context.

- **Pattern syntax:** colon-prefixed named segments. Example: `/users/:id`, `/users/:userId/orders/:orderId`.
- **Scoring:** scores higher than the existing `MatchPath` when a pattern match succeeds (prefer pattern over exact-path wildcard).
- **Config dispatch:** registered as `"type": "path_pattern"` with a `pattern` field in the declarative matcher config.

### Template side

- `{{pathParam.NAME}}` resolves to the captured value for named segment `NAME`.
- `{{request.query.NAME}}` is verified and documented as a first-class accessor.
- `{{request.path}}` continues to work for the raw path.

## Acceptance criteria

- [ ] Parser extended to support function calls with keyword arguments: `{{func arg1=val1 arg2=val2}}`.
- [ ] Built-in helpers: `now`, `uuid`, `randomHex`, `randomInt`, `counter`.
- [ ] `faker.*` bridge to the existing Faker surface (email, name, phone, uuid, address — match all built-in Faker types in `docs/api-reference.md`).
- [ ] Counter is per-server-instance, per-name, survives across requests within the server's lifetime. Reset via admin endpoint or `Server.ResetCounter(name)`.
- [ ] `{{request.*}}` continues to work (backward compatible).
- [ ] Strict mode (`WithStrictTemplating`) continues to reject unresolvable helpers with a 500.
- [ ] Tests covering every helper (unit) and a full template end-to-end (integration).
- [ ] Docs: new `docs/template-helpers.md`, update `docs/replay.md` templating section.
- [ ] `PathPatternCriterion` with `Pattern string` field; `Score` exposes captures via the request's evaluation context.
- [ ] Config dispatch table registers `"path_pattern"`.
- [ ] JSON Schema (`config.schema.json`) enum extended with `"path_pattern"`.
- [ ] `{{pathParam.NAME}}` resolves correctly for all captured segments.
- [ ] `{{request.query.NAME}}` verified and documented as a first-class accessor.
- [ ] Nested template evaluation (template value inside a helper argument, e.g., `seed=user-{{pathParam.id}}`) works correctly.
- [ ] Integration test covering `/users/:id` exemplar-style request with `{{pathParam.id}}` in response template.
- [ ] Docs: `docs/matching.md` updated with path pattern section; `docs/template-helpers.md` documents `pathParam` accessor.

1. **Parser strategy**: The current `{{request.*}}` parser is ad-hoc. Extending it for helpers with keyword args pushes it toward template-engine territory. Incrementally extend, or write a small proper tokenizer? PM leans proper tokenizer — the complexity creep otherwise is painful.
2. **`{{faker.email}}` without a seed**: Use a request-scoped deterministic source (e.g., hash the incoming request) so it is deterministic per request even without an explicit seed? Architect to pin down the fallback policy.
3. **Counter overflow**: Wrap at int64 max? Reset? Error? Low-priority edge case; architect's call.
4. **Pattern syntax**: Express style (`:id`) vs OpenAPI style (`{id}`). PM recommends Express style (more readable in Go source, matches common Go HTTP router idioms). Architect to confirm.
5. **Match precedence vs `MatchPath`**: a `PathPatternCriterion` matching `/users/:id` should score higher than a regex/wildcard `MatchPath` but lower than an exact `MatchPath` match. Architect to specify the exact weight (suggest 3, between `MatchPath=2` and `MatchHeaders=3`).
6. **Query-param-in-pattern support**: out of scope for this issue. Query params stay accessed via `{{request.query.*}}`.

## Dependencies

None directly, but integrates nicely with the request journal feature (#194) for admin reset endpoint for counters. #199 (synthetic responses via exemplar tapes) has a hard dependency on this issue.

## Out of scope

- No user-defined custom helpers via plugin/scripting.
- No template inheritance / partials.
- No conditional logic (`{{#if}}`).
- No iteration (`{{#each}}`).
- No query-param patterns (`?foo=:bar`) — use `{{request.query.*}}` accessors instead.
- Type coercion helpers (`| int`, `| float`, `| bool`) are deferred to #199 where they are needed for exemplar JSON body rendering.

tibtof:
Closes #196

## Summary

- **Template helpers**: `{{now}}`, `{{uuid}}`, `{{randomHex}}`, `{{randomInt}}`, `{{counter}}`, and `{{faker.*}}` (email, name, phone, address, creditCard, hmac, redacted, uuid) with keyword argument support
- **PathPatternCriterion**: Express-style path patterns (e.g., `/users/:id`) compiled to regex at construction time, score 3, with `ExtractParams` for captured segments
- **`{{pathParam.*}}` accessor**: resolves captured path segments from PathPatternCriterion in template expressions
- **Nested template evaluation**: `{{faker.name seed=user-{{pathParam.id}}}}` resolved via balanced delimiter scanning
- **JSON-aware template resolution**: JSON bodies are deserialized, templates in string values resolved, then re-serialized with proper escaping
- **`Server.ResetCounter`**: resets named or all template counters
- **`ResolveTemplateBodySimple`**: backward-compatible wrapper for external callers
- **Config support**: `"type": "path_pattern"` criterion with `"pattern"` field, JSON Schema updated

### Breaking changes (pre-1.0)

- `ResolveTemplateBody` and `ResolveTemplateHeaders` now accept `*templateCtx` (unexported) instead of `*http.Request`
- Unknown template namespaces (e.g., `{{state.counter}}`) now resolve to empty string in lenient mode instead of being left as literal text

## Test plan

- [ ] `go test -race -count=1 ./...` passes (verified locally)
- [ ] `TestParseExpr` -- table-driven for all expression forms
- [ ] `TestResolveNow` -- RFC3339, unix, unixMillis, custom format
- [ ] `TestResolveUUID` -- format, version/variant bits
- [ ] `TestResolveRandomHex` -- length, hex validity, error cases
- [ ] `TestResolveRandomInt` -- range bounds, error cases
- [ ] `TestResolveCounter` -- sequential increment, named independence, wrap at MaxInt64
- [ ] `TestCounterState_Concurrent` -- 1000 goroutines, mutex correctness
- [ ] `TestCounterState_Race` -- dedicated race test in race_test.go
- [ ] `TestResolveFaker_*` -- email, name, phone, address, uuid, auto-seed, unknown type
- [ ] `TestResolvePathParam` -- existing, missing, nil pathParams
- [ ] `TestResolveArgs_Nested` -- `seed=user-{{pathParam.id}}` resolution
- [ ] `TestResolveTemplateJSON` -- string resolution, nested objects, arrays, JSON escaping, strict error
- [ ] `TestPathPatternCriterion_*` -- basic match, multi-segment, no match, exact literal, trailing slash, empty segment, extract params, errors, name, tape must match
- [ ] `TestServer_ResetCounter` -- increment, reset, re-increment
- [ ] `TestServer_PathParamTemplating` -- end-to-end with PathPatternCriterion
- [ ] `TestServer_HelperTemplating` -- counter + faker in response body
- [ ] `TestServer_CounterAcrossRequests` -- counter persists across 3 requests
- [ ] `TestLoadConfig_PathPatternCriterion` -- valid config loading
- [ ] `TestConfig_BuildMatcher_PathPattern` -- build and match
- [ ] `TestLoadConfig_PathPatternValidationErrors` -- 7 error cases
- [ ] `TestIntegration_PathParamFaker` -- end-to-end with nested `{{faker.name seed=user-{{pathParam.id}}}}`

### DIFF
--- a/templating.go
+++ b/templating.go
@@ -313,7 +313,11 @@ func resolveExpr(expr string, ctx *templateCtx) (string, bool) {
 		return resolveNow(args), true
 
 	case pe.name == "uuid":
-		return resolveUUID(ctx), true
+		val, err := resolveUUID(ctx)
+		if err != nil {
+			return "", false
+		}
+		return val, true
 
 	case pe.name == "randomHex":
 		val, err := resolveRandomHex(args, ctx)
@@ -407,14 +411,17 @@ func resolveNow(args map[string]string) string {
 }
 
 // resolveUUID generates a random UUID v4 string using ctx.randSource.
-func resolveUUID(ctx *templateCtx) string {
+// Returns an error if the random source fails.
+func resolveUUID(ctx *templateCtx) (string, error) {
 	var buf [16]byte
-	_, _ = io.ReadFull(ctx.randSource, buf[:])
+	if _, err := io.ReadFull(ctx.randSource, buf[:]); err != nil {
+		return "", fmt.Errorf("resolveUUID: crypto/rand read: %w", err)
+	}
 	// Set version 4 and variant RFC 4122.
 	buf[6] = (buf[6] & 0x0f) | 0x40
 	buf[8] = (buf[8] & 0x3f) | 0x80
 	return fmt.Sprintf("%08x-%04x-%04x-%04x-%012x",
-		buf[0:4], buf[4:6], buf[6:8], buf[8:10], buf[10:16])
+		buf[0:4], buf[4:6], buf[6:8], buf[8:10], buf[10:16]), nil
 }
 
 // resolveRandomHex generates a random hex string of the specified length.
@@ -430,7 +437,9 @@ func resolveRandomHex(args map[string]string, ctx *templateCtx) (string, error) 
 	}
 	nBytes := (length + 1) / 2
 	buf := make([]byte, nBytes)
-	_, _ = io.ReadFull(ctx.randSource, buf)
+	if _, err := io.ReadFull(ctx.randSource, buf); err != nil {
+		return "", fmt.Errorf("randomHex: crypto/rand read: %w", err)
+	}
 	return hex.EncodeToString(buf)[:length], nil
 }
 
@@ -455,10 +464,23 @@ func resolveRandomInt(args map[string]string, ctx *templateCtx) (string, error) 
 	if minVal > maxVal {
 		return "", fmt.Errorf("randomInt min (%d) must be <= max (%d)", minVal, maxVal)
 	}
+	if minVal == maxVal {
+		return strconv.FormatInt(minVal, 10), nil
+	}
 
-	rangeSize := maxVal - minVal + 1
-	n, _ := rand.Int(ctx.randSource, big.NewInt(rangeSize))
-	result := minVal + n.Int64()
+	// Use math/big arithmetic to avoid int64 overflow when the range
+	// (maxVal - minVal + 1) exceeds math.MaxInt64. For example,
+	// min=0, max=math.MaxInt64 would overflow int64 subtraction.
+	hi := new(big.Int).SetInt64(maxVal)
+	lo := new(big.Int).SetInt64(minVal)
+	rangeSize := new(big.Int).Sub(hi, lo)
+	rangeSize.Add(rangeSize, big.NewInt(1))
+
+	n, err := rand.Int(ctx.randSource, rangeSize)
+	if err != nil {
+		return "", fmt.Errorf("randomInt: crypto/rand: %w", err)
+	}
+	result := new(big.Int).Add(n, lo).Int64()
 	return strconv.FormatInt(result, 10), nil
 }
 
@@ -704,6 +726,9 @@ func resolveTemplateBytes(body []byte, ctx *templateCtx, strict bool) ([]byte, e
 //
 // Headers that contain no "{{" sequences are copied as-is (fast path per
 // header value).
+//
+// For callers outside the package, use [ResolveTemplateHeadersSimple] which
+// constructs the evaluation context from an *http.Request.
 func ResolveTemplateHeaders(h http.Header, ctx *templateCtx, strict bool) (http.Header, error) {
 	if h == nil {
 		return nil, nil
@@ -765,6 +790,8 @@ func resolveTemplateStringCtx(s string, ctx *templateCtx, strict bool) (string, 
 // ResolveTemplateBodySimple is a backward-compatible convenience wrapper that
 // resolves templates using only request data (no path params, no counters,
 // no faker). Equivalent to the pre-#196 ResolveTemplateBody behavior.
+//
+// See also [ResolveTemplateHeadersSimple] for the header equivalent.
 func ResolveTemplateBodySimple(body []byte, r *http.Request, strict bool) ([]byte, error) {
 	if !bytes.Contains(body, []byte("{{")) {
 		return body, nil
@@ -777,6 +804,25 @@ func ResolveTemplateBodySimple(body []byte, r *http.Request, strict bool) ([]byt
 		randSource: rand.Reader,
 	}
 	return ResolveTemplateBody(body, ctx, strict)
+}
+
+// ResolveTemplateHeadersSimple is a convenience wrapper that resolves
+// templates in header values using only request data (no path params,
+// no counters, no faker). It constructs the unexported evaluation context
+// internally, making it callable from outside the package.
+//
+// See also [ResolveTemplateBodySimple] for the body equivalent.
+func ResolveTemplateHeadersSimple(headers http.Header, r *http.Request, strict bool) (http.Header, error) {
+	if headers == nil {
+		return nil, nil
+	}
+	reqBody := readRequestBody(r)
+	ctx := &templateCtx{
+		req:        r,
+		reqBody:    reqBody,
+		randSource: rand.Reader,
+	}
+	return ResolveTemplateHeaders(headers, ctx, strict)
 }
 
 // readRequestBody reads the full request body and restores it so the body
--- a/templating.go
+++ b/templating.go
@@ -2,11 +2,19 @@ package httptape
 
 import (
 	"bytes"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
+	"math/big"
 	"net/http"
 	"strconv"
+	"strings"
+	"sync"
+	"time"
 )
 
 // templateExpr represents a parsed template expression extracted from a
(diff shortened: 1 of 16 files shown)


templating.go
```diff
@@ -107,14 +376,158 @@ func resolveExpr(expr string, r *http.Request, reqBody []byte) (string, bool) {
...
+		maxVal, err = strconv.ParseInt(s, 10, 64)
+		if err != nil {
+			return "", fmt.Errorf("randomInt \"max\" must be an integer, got %q", s)
+		}
+	}
+	if minVal > maxVal {
+		return "", fmt.Errorf("randomInt min (%d) must be <= max (%d)", minVal, maxVal)
+	}
+
+	rangeSize := maxVal - minVal + 1
```
tibtof: **Panic risk**: `rangeSize := maxVal - minVal + 1` can overflow `int64` when the range exceeds `math.MaxInt64` (e.g., `min=0 max=9223372036854775807`). The overflow produces a negative value, which causes `rand.Int(ctx.randSource, big.NewInt(rangeSize))` to panic (`rand.Int` panics when max <= 0). A user-controlled template expression like `{{randomInt min=0 max=9223372036854775807}}` triggers this.

This violates the project rule: no `panic` in library hot paths -- a library must never panic on user input.

**Fix**: Use `math/big` arithmetic to compute the range safely:
```go
bMin := big.NewInt(minVal)
bMax := big.NewInt(maxVal)
rangeSize := new(big.Int).Sub(bMax, bMin)
rangeSize.Add(rangeSize, big.NewInt(1))
if rangeSize.Sign() <= 0 {
    return "", fmt.Errorf("randomInt range overflow")
}
n, err := rand.Int(ctx.randSource, rangeSize)
if err != nil {
    return "", fmt.Errorf("randomInt: reading random source: %w", err)
}
result := new(big.Int).Add(bMin, n)
return result.String(), nil
```

templating.go
```diff
@@ -107,14 +376,158 @@ func resolveExpr(expr string, r *http.Request, reqBody []byte) (string, bool) {
...
+	default:
+		// Treat as a custom Go time format string.
+		return t.Format(format)
+	}
+}
+
+// resolveUUID generates a random UUID v4 string using ctx.randSource.
+func resolveUUID(ctx *templateCtx) string {
+	var buf [16]byte
+	_, _ = io.ReadFull(ctx.randSource, buf[:])
```
tibtof: **Swallowed errors**: `io.ReadFull` errors are discarded with `_, _ =`. While `crypto/rand.Reader` is extremely unlikely to fail, the project code style rule is "never swallow errors." The same pattern appears in `resolveRandomHex` (line 433) and `rand.Int` in `resolveRandomInt` (line 460).

I recognize the existing `newUUID()` in `tape.go` follows the same pattern, so there is project precedent. **Non-blocking** -- but worth adding a comment (e.g., `// crypto/rand.Reader never fails in practice; ignoring error per newUUID precedent`) to make the intent explicit.

templating.go
```diff
@@ -213,39 +689,32 @@ func ResolveTemplateBody(body []byte, r *http.Request, strict bool) ([]byte, err
...
 //
-// Non-"request." expressions are left as literal text (forward-compat).
-//
 // Headers that contain no "{{" sequences are copied as-is (fast path per
 // header value).
-//
-// ResolveTemplateHeaders is safe for concurrent use — it is a pure function
-// with no shared mutable state.
-func ResolveTemplateHeaders(h http.Header, r *http.Request, strict bool) (http.Header, error) {
+func ResolveTemplateHeaders(h http.Header, ctx *templateCtx, strict bool) (http.Header, error) {
```
tibtof, line 732: **API design issue**: `ResolveTemplateHeaders` is exported but its second parameter is `*templateCtx` (unexported type). External callers cannot construct this argument, making the function uncallable from outside the package. `ResolveTemplateBody` has the same issue but gets a `ResolveTemplateBodySimple` wrapper. There is no equivalent `ResolveTemplateHeadersSimple`.

**Options** (pick one):
1. Add `ResolveTemplateHeadersSimple(h http.Header, r *http.Request, strict bool) (http.Header, error)` for symmetry.
2. Unexport `ResolveTemplateHeaders` (rename to `resolveTemplateHeaders`) since external callers can't use it anyway.

I recommend option 2 since the function is only called from `server.go`'s `ServeHTTP`. If a user ever needs it, option 1 can be added later.

**Non-blocking** -- this is a pre-1.0 library so the API surface can be adjusted. But leaving an exported-but-uncallable function is confusing for godoc readers.

tibtof (commented):
## Review Summary — PR #235 (feat: template helpers, PathPatternCriterion, JSON-aware resolution)

### Verdict: REQUEST CHANGES (1 blocker)

### Blocker

**`resolveRandomInt` integer overflow causes panic** (`templating.go:459`). When `max - min + 1` exceeds `math.MaxInt64`, the `int64` arithmetic overflows to a negative value, and `crypto/rand.Int` panics on non-positive max. A user-controlled template expression like `{{randomInt min=0 max=9223372036854775807}}` triggers this. The project constitution forbids panics in library hot paths. Fix: use `math/big` arithmetic for the range calculation (see inline comment for concrete code).

### Non-blocking observations (2)

1. **Swallowed `io.ReadFull` / `rand.Int` errors** (`templating.go:412, 433, 460`): consistent with existing `newUUID()` precedent in `tape.go`, but worth a justification comment.

2. **`ResolveTemplateHeaders` exported with unexported `*templateCtx` param** (`templating.go:707`): uncallable from outside the package, no `Simple` wrapper provided (unlike `ResolveTemplateBody`). Recommend either unexporting or adding a wrapper.

### What passed review

- **ADR-42 fidelity**: all 3 spot-checked design points match (Express-style `:id` syntax, score 3 for PathPatternCriterion, auto-seed = SHA-256 of tapeID + ":" + path).
- **Stdlib only**: no external dependencies added. `go.mod` is clean.
- **Backward compatibility**: `ResolveTemplateBodySimple` provides the old API. CHANGELOG documents both breaking changes (signature change + unknown-namespace behavior change).
- **Counter thread-safety**: `counterState.mu` protects all reads/writes. Race test (1000 goroutines) passes under `-race`.
- **PathPatternCriterion correctness**: trailing slash, empty segment, multi-param, no-match, tape-must-match, malformed patterns — all tested.
- **Config validation**: all existing criterion validators reject `Pattern` field; `path_pattern` rejects `Paths`.
- **JSON Schema**: `path_pattern` enum, `pattern` property, `if/then` constraints — all consistent.
- **Build/vet/race/fmt**: `go build`, `go vet`, `go test -race` all pass. `gofmt` reports only pre-existing formatting diffs (unrelated files).
- **Architecture**: single flat package, no sub-packages, no `init()`, no global mutable state, no panics (except the blocker above).
- **Test coverage**: comprehensive. Table-driven tests for `parseExpr`, all helpers, path params, nested template evaluation, counter wrap, JSON-aware resolution, backward-compat wrapper. Integration test with nested faker seed. Race test for counter state.
- **Docs**: `docs/template-helpers.md` examples match implementation. Updated `matching.md`, `api-reference.md`, `replay.md`.

### Breaking changes (2, both justified by ADR-42)

1. **`ResolveTemplateBody` / `ResolveTemplateHeaders` signature change**: second param changed from `*http.Request` to `*templateCtx` (unexported). ADR-42 explicitly calls this out as acceptable pre-1.0 with `ResolveTemplateBodySimple` as migration path.

2. **Unknown template namespaces**: `{{state.counter}}` previously left as literal text, now replaced with empty string (lenient) or error (strict). ADR-42 mandates explicit dispatch with no fallthrough. CHANGELOG documents this.

tibtof:
## Review feedback addressed (3/3)

All three items from the review have been fixed in commit 00c2f1f.

### 1. BLOCKER -- `resolveRandomInt` int64 overflow

`resolveRandomInt` now uses `math/big` arithmetic for the range calculation, preventing overflow when `maxVal - minVal + 1` exceeds `math.MaxInt64`. Added early return for the `min == max` case (avoids `rand.Int` with range 1).

Regression tests added:
- `min=0, max=MaxInt64` -- produces a non-negative int64 without panic
- `min=MinInt64, max=MaxInt64` -- full int64 range, no panic
- `min=5, max=5` -- returns exactly 5

All pass under `-race -count=10` with zero flakes.

### 2. Non-blocker -- swallowed `crypto/rand` errors

**Decision: converted to error propagation** rather than swallow-with-comment. Rationale:

- `resolveRandomHex` and `resolveRandomInt` already returned `(string, error)`, so propagating the `io.ReadFull` / `rand.Int` error required zero signature changes -- just wrapping and returning.
- `resolveUUID` returned only `string`. I changed it to `(string, error)` and updated the single call site in `resolveExpr` to handle the error (returns `"", false`, same as other unresolvable expressions). This is an unexported function so no external API breakage.
- The `newUUID()` precedent in `tape.go` already handles the error (returns a zero UUID), so the project convention is *not* to swallow -- it is to handle. The `_, _ =` pattern in the original code was inconsistent with `tape.go`.

### 3. Non-blocker -- `ResolveTemplateHeadersSimple` wrapper

Added `ResolveTemplateHeadersSimple(headers http.Header, r *http.Request, strict bool) (http.Header, error)` that constructs the unexported `*templateCtx` internally, paralleling `ResolveTemplateBodySimple`. Updated godoc on both `ResolveTemplateHeaders` and `ResolveTemplateBodySimple` to cross-reference their `Simple` counterparts.

Tests added for the new wrapper (template resolution and nil-header handling).

---

Build/vet/test all pass:
```
go build ./...  -- OK
go vet ./...    -- OK
go test -race -count=1 ./...  -- OK (21s)
gofmt -l templating.go templating_test.go  -- no output (clean)
```

Ready for re-review.

tibtof (commented):
LGTM — APPROVE (submitted as comment because GitHub prevents self-approval).

Re-review of commit 00c2f1f — all three prior review items verified as fixed.

**1. Blocker fix (int64 overflow in `resolveRandomInt`): CONFIRMED FIXED.**
The `math/big` arithmetic is correct: `rangeSize = hi - lo + 1` computed in arbitrary precision, `rand.Int(rng, rangeSize)` returns `n` in `[0, rangeSize)`, and `result = n + lo` always lands in `[minVal, maxVal]` which is within int64 range by construction. The `min == max` early return avoids unnecessary crypto/rand calls. The `min > max` validation returns an error (no panic). Regression tests cover `min=0/max=MaxInt64`, `min=MinInt64/max=MaxInt64`, and `min=max=5` — all pass under `-race -count=10`.

**2. Swallowed errors: CONFIRMED FIXED.**
`resolveRandomHex` and `resolveRandomInt` now propagate `io.ReadFull`/`rand.Int` errors with proper `fmt.Errorf` wrapping. `resolveUUID` signature changed from `string` to `(string, error)` — unexported (`r` lowercase), single call site in `resolveExpr` updated to handle the error consistently with other helpers (`return "", false`). No exported API impact.

**3. `ResolveTemplateHeadersSimple` wrapper: CONFIRMED FIXED.**
Signature `(headers http.Header, r *http.Request, strict bool) (http.Header, error)` mirrors `ResolveTemplateBodySimple`. Godoc cross-references present on both `ResolveTemplateHeaders` (points to `Simple` variant) and `ResolveTemplateBodySimple` (points to `HeadersSimple`). Tests cover template resolution and nil-header edge case.

**Build verification:**
- `go test -race -count=10 -run "RandomInt" ./...` — PASS
- `go test -race -count=1 ./...` — PASS (21s)
- `go vet ./...` — clean
- `gofmt -l .` — only pre-existing files (not part of this PR)

No new issues found. Ready for human review.

PR #236, opened 2026-04-26
feat(#199): synthetic responses via exemplar tapes

### LINKED ISSUE #199: feat: synthetic responses via exemplar tapes (opt-in, frontend-first workflows)
## Problem statement

Today httptape is strict "replay recorded traffic." A request that doesn't match any tape falls through to the fallback handler (404 or configured). This is correct for integration tests — unexpected requests should fail fast so real bugs aren't masked. But in frontend-first development workflows — where the backend is partially specified and the UI is being built against a mock — this is too strict. A developer would record one tape for `/users/1` and want the server to synthesize `/users/2`, `/users/3`, etc. with realistic fake data, all deterministic so the UI behaves reproducibly across reloads.

## User story

As a frontend developer building against a partially-specified backend, I want to mark certain tapes as "exemplars" with URL patterns and template bodies so that the mock server synthesizes realistic, deterministic responses for any matching URL — without requiring a separate tape for every possible parameter value.

### Per-tape flag

Tapes can be marked as "exemplars" — they define a URL pattern (e.g., `/users/:id`) and a template body. When the server is running in synthesis mode AND no exact match is found, exemplars are consulted; the first matching exemplar's body is rendered with captured path params, the deterministic faker, and the other template helpers.

```json
{
  "id": "users-exemplar",
  "route": "users-by-id",
  "request": {
    "method": "GET",
    "url_pattern": "/users/:id"
  },
  "exemplar": true,
  "response": {
    "status_code": 200,
    "headers": {"Content-Type": ["application/json"]},
    "body": {
      "id": "{{pathParam.id | int}}",
      "name": "{{faker.name seed=user-{{pathParam.id}}}}",
      "email": "{{faker.email seed=user-{{pathParam.id}}}}",
      "createdAt": "{{now}}"
    }
  }
}
```

- New optional field `exemplar: bool` on the tape struct (default false).
- New optional field `url_pattern: string` on the `request` struct (mutually exclusive with exact `url`).
- When `exemplar: true`, the request URL is the pattern.

### Critical design constraint: opt-in at the server level

Synthesis must be explicitly enabled at server startup. When off, exemplar tapes are loaded but ignored (no pattern fallback). This protects integration tests from silently masking missing fixtures. The per-tape `exemplar: true` field is a necessary but NOT sufficient condition for synthesis — the server-level flag is the master switch.

### Matching precedence (when synthesis is enabled)

1. Exact URL match on a non-exemplar tape wins.
2. If no exact match, exemplar tapes with matching `url_pattern` are considered.
3. Among multiple matching exemplars: most specific wins (fewest wildcard segments); ties broken by order of declaration.

### Matching precedence (when synthesis is disabled)

1. Exact URL match on a non-exemplar tape wins.
2. Exemplar tapes are ignored entirely — fallback handler runs as today.

### Server-level opt-in

- **CLI flag:** `httptape serve --synthesize`
- **ServerOption:** `WithSynthesis() ServerOption`
- **Startup logging:** clearly indicates synthesis state, e.g., `httptape: synthesis mode ENABLED — N exemplar tapes loaded`

### Template resolution in native JSON bodies

In the polymorphic body model (object / string / base64), template resolution for exemplars must handle all forms:

- **Native JSON objects:** recurse into object values. Strings at leaf positions get template-resolved. Integers / booleans / nulls stay as-is unless they are string-encoded templates like `"{{pathParam.id | int}}"`.
- **String bodies:** template-resolve the string directly.
- **Base64 bodies:** no template resolution (binary data).

### Type coercion helpers

Templates produce strings by default. For native JSON body rendering where type correctness matters, coercion helpers are needed:

- `| int` — coerce template result to JSON integer
- `| float` — coerce template result to JSON float
- `| bool` — coerce template result to JSON boolean

## Acceptance criteria

- [ ] `exemplar: bool` field on `Tape` struct (optional, default false, backward compatible).
- [ ] `url_pattern: string` field on `RecordedReq` struct (optional, mutually exclusive with exact `url`).
- [ ] Validation: tape with `exemplar: true` must have `url_pattern` set; tape with exact `url` must not have `url_pattern`.
- [ ] `PathPatternCriterion` (from #196) drives exemplar matching.
- [ ] `WithSynthesis()` ServerOption and `--synthesize` CLI flag on `serve`.
- [ ] Precedence: exact URL wins over exemplar; among exemplars, most-specific wins.
- [ ] Template body resolution: recursive on native JSON objects, string-level on string bodies, no-op on base64.
- [ ] Type coercion helpers in templates: `| int`, `| float`, `| bool`.
- [ ] Startup logging clearly indicates synthesis state.
- [ ] When synthesis off: exemplar tapes loaded but never selected (integration tests safe).
- [ ] Round-trip property on exemplar tapes: `Marshal(Unmarshal(json)) == json`.
- [ ] Unit tests: precedence logic, pattern matching, template resolution, coercion.
- [ ] Integration test: exemplar tape serving multiple unique IDs (`/users/1`, `/users/2`, `/users/7`) with deterministic distinct fake data per ID.
- [ ] Negative test: synthesis disabled, exemplar ignored, fallback 404.
- [ ] Docs: new `docs/synthesis.md` (when to use, how it differs from replay, opt-in gating), update `docs/fixtures-authoring.md`, `docs/cli.md`, `docs/api-reference.md`.
- [ ] CHANGELOG entry describing the new mode.

1. **`url_pattern` as separate field vs overloading `url`:** PM recommends separate — cleaner typing and unambiguous validation. Architect to confirm.
2. **Can a single tape be both exact-url and exemplar?** No — mutually exclusive per validation rule above. But a tape library can have multiple tapes covering the same route (one exact for `/users/1`, one exemplar for `/users/:id`).
3. **Synthesis + SSE:** can exemplar tapes serve SSE streams? The template must operate on each `sse_event` body. Architect to decide scope (probably: SSE exemplars are out of scope; file follow-up if needed).
4. **Synthesis + request body matching:** `body_fuzzy` criterion requires a request body to compare against. Exemplars are typically GET-shaped. If a POST exemplar is declared, how does `body_fuzzy` interact? PM suggests: exemplars with POST/PUT/etc. are allowed but only consulted when there is no exact match AND the matcher criteria (including `body_fuzzy`) also pass on the exemplar's request.
5. **Persistence:** should exemplars live in the same directory as normal tapes (just a flag on the JSON) or in a subdirectory? PM recommends same directory, flag-based discrimination. Architect to confirm.
6. **Faker seed stability across exemplar redeclarations:** if the user renames a tape file or edits the pattern, the seed `user-{{pathParam.id}}` still produces the same fake (because it is seed-content-driven, not tape-identity-driven). Document as a feature.

## Dependencies

- **Hard dependency on #196** (template helpers + path-param accessors + `PathPatternCriterion`). This issue cannot ship before #196.
- Independent of #194 (journal), #195 (scenarios), #197 (inbound TLS).

## Out of scope

- Synthesis for SSE fixtures (file follow-up if needed).
- Schema-based synthesis (learn a contract, synthesize anything valid) — too ambitious for this issue.
- Stateful exemplars (one exemplar that remembers prior calls and adjusts) — that is scenarios territory (#195).
- Query-param patterns (`?foo=:bar`) — stay with exact match + `{{request.query.*}}` accessors.
- Response generation via ML or schema inference.
- User-defined custom helpers or scripting.

tibtof:
Closes #199

## Summary

Implements synthesis mode (ADR-43) -- exemplar tapes with URL patterns and template expressions generate responses for unmatched URLs.

**New public API surface:**
- `Tape.Exemplar bool` -- marks a tape as a pattern-based template
- `RecordedReq.URLPattern string` -- colon-prefixed path pattern (e.g., `/users/:id`)
- `WithSynthesis() ServerOption` -- enables exemplar fallback on the Server
- `ValidateTape(t Tape) error` -- structural validation (calls ValidateExemplar)
- `ValidateExemplar(t Tape) error` -- exemplar-specific constraint checks
- `--synthesize` CLI flag for `httptape serve`

**Type coercion in JSON bodies:**
- `| int`, `| float`, `| bool` pipes convert resolved template strings to native JSON types
- Only effective in JSON response bodies of exemplar tapes
- Strict mode: coercion failure returns HTTP 500; lenient mode: emits uncoerced string

**Two-phase match flow:**
1. Phase 1: exact tapes checked first via existing matcher
2. Phase 2: exemplar fallback (only when synthesis enabled and phase 1 misses)
3. Exemplars must pass all configured criteria (method, headers, body_fuzzy, etc.)

**Opt-in design (fail-safe defaults):**
- Synthesis disabled by default -- integration tests are unaffected
- Exemplar tapes are loaded but never consulted unless `WithSynthesis()` is set
- CLI warns when exemplar tapes exist but synthesis is disabled

**Startup validation:**
- CLI validates all loaded tapes at startup
- Invalid exemplar tapes (missing url_pattern, SSE exemplar, etc.) cause startup error

## Test plan

- [x] `TestValidateExemplar` -- all 7 validation rules (valid exemplar, missing url_pattern, mutual exclusivity, SSE rejection, etc.)
- [x] `TestValidateTape_DelegatesToValidateExemplar` -- ValidateTape dispatches to ValidateExemplar
- [x] `TestTape_MarshalJSON_Exemplar` -- exemplar fields appear in JSON output
- [x] `TestTape_UnmarshalJSON_Exemplar` -- exemplar fields parsed correctly
- [x] `TestTape_UnmarshalJSON_NoExemplar` -- backward compatibility (no exemplar fields)
- [x] `TestTape_MarshalJSON_RoundTrip_Exemplar` -- marshal/unmarshal/re-marshal identity
- [x] `TestServer_SynthesisDisabled_ExemplarIgnored` -- 404 when synthesis off
- [x] `TestServer_SynthesisEnabled_ExactWins` -- exact tape beats exemplar
- [x] `TestServer_SynthesisEnabled_ExemplarFallback` -- exemplar consulted on miss
- [x] `TestServer_SynthesisEnabled_MostSpecificExemplar` -- /users/:id/orders wins over /users/:id
- [x] `TestServer_SynthesisEnabled_MethodFilter` -- POST doesn't match GET exemplar
- [x] `TestServer_SynthesisEnabled_NoExemplars` -- 404 when no exemplars exist
- [x] `TestServer_SynthesisEnabled_DeterministicResponse` -- same request = same output
- [x] `TestServer_SynthesisEnabled_ExemplarWithTextBody` -- text/plain body resolution
- [x] `TestServer_SynthesisEnabled_PatternWithNoParams` -- exact path exemplar works
- [x] `TestParseCoercion` -- 7 cases including edge cases
- [x] `TestCoerceValue` -- int, float, bool, error, whitespace handling
- [x] `TestResolveTemplateJSONWithCoercion_*` -- 10 tests covering objects, arrays, nesting, coercion, mixed content, strict/lenient failure
- [x] `TestResolveExemplarBody_*` -- JSON, text, binary, empty body dispatch
- [x] `TestIntegration_ExemplarSynthesis` -- end-to-end with httptest.NewServer
- [x] `TestIntegration_SynthesisDisabled_ExemplarIgnored` -- end-to-end negative case
- [x] `TestServeWithSynthesize` -- CLI flag acceptance
- [x] `TestServeWithSynthesizeLogging` -- CLI with exemplar fixture
- [x] `TestServeWithInvalidExemplarTape` -- startup validation rejects bad exemplar
- [x] All existing tests pass (no regressions)
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -race -count=1 ./...` clean
- [x] `gofmt -l .` clean (only pre-existing issues in unmodified files)

### DIFF
--- a/cmd/httptape/main.go
+++ b/cmd/httptape/main.go
@@ -251,17 +251,14 @@ func runServe(args []string) error {
 		}
 		logger.Printf("synthesis mode ENABLED -- %d exemplar tape(s) loaded", exemplarCount)
 	} else {
+		exemplarCount := 0
 		for _, t := range allTapes {
 			if t.Exemplar {
-				exemplarCount := 0
-				for _, tt := range allTapes {
-					if tt.Exemplar {
-						exemplarCount++
-					}
-				}
-				logger.Printf("WARNING: %d exemplar tape(s) found but synthesis is disabled (use --synthesize to enable)", exemplarCount)
-				break
+				exemplarCount++
 			}
+		}
+		if exemplarCount > 0 {
+			logger.Printf("WARNING: %d exemplar tape(s) found but synthesis is disabled (use --synthesize to enable)", exemplarCount)
 		}
 	}
 
--- a/cmd/httptape/main.go
+++ b/cmd/httptape/main.go
@@ -174,6 +174,7 @@ func runServe(args []string) error {
 	var replayHeaders repeatableFlag
 	fs.Var(&replayHeaders, "replay-header", "Header to inject into responses (Key=Value, repeatable)")
 	sseTiming := fs.String("sse-timing", "", "SSE replay timing mode: realtime, instant, accelerated=<factor>")
+	synthesize := fs.Bool("synthesize", false, "Enable synthesis mode (exemplar tapes generate responses for unmatched URLs)")
 
 	if err := fs.Parse(args); err != nil {
 		return &usageError{err}
@@ -224,6 +225,44 @@ func runServe(args []string) error {
 			return &usageError{fmt.Errorf("invalid --replay-header %q: expected Key=Value", rh)}
 		}
 		serverOpts = append(serverOpts, httptape.WithReplayHeaders(rh[:eqIdx], rh[eqIdx+1:]))
+	}
+	if *synthesize {
+		serverOpts = append(serverOpts, httptape.WithSynthesis())
+	}
+
+	// Startup validation: check all loaded tapes for structural validity.
+	allTapes, err := store.List(context.Background(), httptape.Filter{})
+	if err != nil {
+		return fmt.Errorf("load tapes: %w", err)
+	}
+	for _, t := range allTapes {
+		if err := httptape.ValidateTape(t); err != nil {
+			return fmt.Errorf("invalid tape %s: %w", t.ID, err)
+		}
+	}
+
+	// Synthesis logging.
+	if *synthesize {
+		exemplarCount := 0
+		for _, t := range allTapes {
+			if t.Exemplar {
+				exemplarCount++
+			}
+		}
+		logger.Printf("synthesis mode ENABLED -- %d exemplar tape(s) loaded", exemplarCount)
+	} else {
+		for _, t := range allTapes {
+			if t.Exemplar {
+				exemplarCount := 0
+				for _, tt := range allTapes {
+					if tt.Exemplar {
+						exemplarCount++
+					}
+				}
+				logger.Printf("WARNING: %d exemplar tape(s) found but synthesis is disabled (use --synthesize to enable)", exemplarCount)
+				break
+			}
+		}
 	}
 
 	server, err := httptape.NewServer(store, serverOpts...)
--- a/cmd/httptape/main_test.go
+++ b/cmd/httptape/main_test.go
@@ -923,12 +923,9 @@ func TestServeWithSynthesizeLogging(t *testing.T) {
 func TestServeWithInvalidExemplarTape(t *testing.T) {
 	// Create a fixture with an invalid exemplar tape (Exemplar=true but no URLPattern).
 	dir := t.TempDir()
-	store, err := httptape.NewFileStore(httptape.WithDirectory(dir))
-	if err != nil {
-		t.Fatalf("create store: %v", err)
-	}
 
-	// Write a malformed exemplar tape directly as JSON.
+	// Write a malformed exemplar tape directly as JSON (bypasses Store validation
+	// to test the startup path's own ValidateTape call).
 	tapeJSON := `{
 		"id": "bad-exemplar",
 		"route": "",
@@ -947,7 +944,6 @@ func TestServeWithInvalidExemplarTape(t *testing.T) {
 			"body": null
 		}
 	}`
-	_ = store
 	tapePath := filepath.Join(dir, "bad-exemplar.json")
 	if err := os.WriteFile(tapePath, []byte(tapeJSON), 0644); err != nil {
 		t.Fatalf("write tape: %v", err)
--- a/cmd/httptape/main_test.go
+++ b/cmd/httptape/main_test.go
@@ -879,6 +879,88 @@ func TestMigrateFixtures_AlreadyMigratedTape(t *testing.T) {
 	}
 }
 
+func TestServeWithSynthesize(t *testing.T) {
+	// --synthesize flag should be accepted without error.
+	got := run([]string{"serve", "--fixtures", t.TempDir(), "--synthesize", "-h"})
+	if got != exitOK {
+		t.Errorf("got exit %d, want %d for serve --synthesize -h", got, exitOK)
+	}
+}
+
+func TestServeWithSynthesizeLogging(t *testing.T) {
+	// Create a fixture directory with an exemplar tape.
+	dir := t.TempDir()
+	store, err := httptape.NewFileStore(httptape.WithDirectory(dir))
+	if err != nil {
+		t.Fatalf("create store: %v", err)
+	}
+
+	exemplar := httptape.Tape{
+		ID:       "exemplar-log-test",
+		Exemplar: true,
+		Request: httptape.RecordedReq{
+			Method:     "GET",
+			URLPattern: "/users/:id",
+		},
+		Response: httptape.RecordedResp{
+			StatusCode: 200,
+			Headers:    http.Header{"Content-Type": {"application/json"}},
+			Body:       []byte(`{"id":"{{pathParam.id}}"}`),
+		},
+	}
+	ctx := context.Background()
+	if err := store.Save(ctx, exemplar); err != nil {
+		t.Fatalf("save exemplar: %v", err)
+	}
+
+	// Just verify --synthesize flag is accepted (the -h flag triggers help and exits).
+	got := run([]string{"serve", "--fixtures", dir, "--synthesize", "-h"})
+	if got != exitOK {
+		t.Errorf("got exit %d, want %d", got, exitOK)
+	}
+}
+
+func TestServeWithInvalidExemplarTape(t *testing.T) {
+	// Create a fixture with an invalid exemplar tape (Exemplar=true but no URLPattern).
+	dir := t.TempDir()
+	store, err := httptape.NewFileStore(httptape.WithDirectory(dir))
+	if err != nil {
+		t.Fatalf("create store: %v", err)
+	}
+
+	// Write a malformed exemplar tape directly as JSON.
+	tapeJSON := `{
+		"id": "bad-exemplar",
+		"route": "",
+		"recorded_at": "2026-01-01T00:00:00Z",
+		"exemplar": true,
+		"request": {
+			"method": "GET",
+			"url": "",
+			"headers": {},
+			"body": null,
+			"body_hash": ""
+		},
+		"response": {
+			"status_code": 200,
+			"headers": {},
+			"body": null
+		}
+	}`
+	_ = store
+	tapePath := filepath.Join(dir, "bad-exemplar.json")
+	if err := os.WriteFile(tapePath, []byte(tapeJSON), 0644); err != nil {
+		t.Fatalf("write tape: %v", err)
+	}
+
+	// The serve command should fail at startup validation because the tape
+	// has Exemplar=true but no URLPattern.
+	got := run([]string{"serve", "--fixtures", dir})
+	if got != exitRuntime {
+		t.Errorf("got exit %d, want %d (startup validation should reject bad exemplar)", got, exitRuntime)
+	}
+}
+
 func itoa(i int) string {
 	const digits = "0123456789"
 	if i == 0 {
--- a/server.go
+++ b/server.go
@@ -504,10 +504,10 @@ func (s *Server) matchExemplar(r *http.Request, exemplars []Tape) (Tape, map[str
 			params = map[string]string{}
 		}
 
-		// Score the path match using the PathPatternCriterion's score logic.
-		// We create a temporary tape with the URL set to the request path so
-		// that the criterion's Score method can compare.
-		pathScore := 3 // PathPatternCriterion always returns 3 on match
+		// PathPatternCriterion's Score returns a flat 3 on match. Same-depth
+		// literal-vs-parameterized exemplars therefore tie and fall back to
+		// declaration order (see follow-up issue for true specificity sort).
+		pathScore := 3
 
 		// Run other criteria against the exemplar. The exemplar must pass all.
 		totalScore := pathScore
--- a/server.go
+++ b/server.go
@@ -33,6 +33,7 @@ type Server struct {
 	sseTiming        SSETimingMode       // controls SSE replay inter-event timing
 	counters         *counterState       // per-server counter state for {{counter}}
 	randSource       io.Reader           // randomness source for template helpers
+	synthesis        bool                // if true, exemplar tapes are consulted on miss
 }
 
 // ServerOption configures a Server.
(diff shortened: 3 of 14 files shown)


server.go
```diff
@@ -391,24 +434,112 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				eliminated = true
+				break
+			}
+			totalScore += score
+		}
+		if eliminated {
+			continue
+		}
+
+		if best == nil || totalScore > best.score || (totalScore == best.score && i < best.declOrder) {
```
tibtof, line 527: **Specificity gap**: when two exemplar patterns match at the same path depth -- e.g., `/users/me` (literal) vs `/users/:id` (parameterized) -- both receive `pathScore = 3` and the tie-breaker is declaration order, not specificity. The ADR anticipated this: *"If #196 uses a flat score, the 'most specific exemplar wins' requirement is met by extending the criterion or implementing a secondary sort here."*

In practice, this is unlikely to bite because a literal-path exemplar (no params) is unusual -- the `/users/me` case should normally be an exact tape, not an exemplar. And patterns at different depths (like the test's `/users/:id` vs `/users/:id/orders`) are already disambiguated by regex anchoring.

**Suggestion (non-blocking)**: add a secondary sort that counts literal segments (more literals = higher specificity). Something like:

```go
// Secondary specificity: fewer param segments = more specific.
specificity := len(strings.Split(exemplar.Request.URLPattern, "/")) - len(ppc.paramNames)
```

Then compare `specificity` when `totalScore` ties, before falling back to declaration order. A follow-up is fine.

server.go
```diff
@@ -391,24 +434,112 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			default:
+				otherCriteria = append(otherCriteria, c)
+			}
+		}
+	}
+
+	for i, exemplar := range exemplars {
+		ppc, err := NewPathPatternCriterion(exemplar.Request.URLPattern)
+		if err != nil {
+			continue // invalid pattern: skip silently
```
tibtof, line 489: **ADR mismatch**: ADR-43 error cases table says invalid URLPattern syntax should be handled as *"Server logs and skips the exemplar"*, but the implementation here skips silently with no log output. Consider logging the error at debug/info level so operators can diagnose typos in exemplar patterns. Example:

```go
if err != nil {
    // Log if a logger is available (or use the server's existing logging mechanism).
    continue
}
```

Non-blocking for this PR since the Server struct currently has no logger field, but worth a follow-up or a TODO comment.

tape.go
```diff
@@ -359,6 +383,53 @@ func NewTape(route string, req RecordedReq, resp RecordedResp) Tape {
...
+
+	if hasURL && hasPattern {
+		return fmt.Errorf("httptape: tape %s: url and url_pattern are mutually exclusive", t.ID)
+	}
+
+	if t.Exemplar {
+		if !hasPattern {
+			return fmt.Errorf("httptape: exemplar tape %s: url_pattern is required", t.ID)
+		}
+		if hasURL {
```
tibtof: **Dead code**: this `if hasURL` check is unreachable. By the time execution reaches line 418, the guard at line 410 (`hasURL && hasPattern`) has already returned. If `t.Exemplar` is true and `hasPattern` is true (line 415 passed), then `hasURL` must be false (because line 410 ensures they can't both be true).

You can safely remove lines 418-420. Not a correctness issue -- just dead code that confuses readers.

cmd/httptape/main.go
```diff
@@ -225,6 +226,44 @@ func runServe(args []string) error {
...
+		for _, t := range allTapes {
+			if t.Exemplar {
+				exemplarCount++
+			}
+		}
+		logger.Printf("synthesis mode ENABLED -- %d exemplar tape(s) loaded", exemplarCount)
+	} else {
+		for _, t := range allTapes {
+			if t.Exemplar {
+				exemplarCount := 0
```
tibtof: **Code quality -- O(n^2) counting**: the else branch iterates `allTapes` to find the first exemplar, then iterates `allTapes` again (inner loop) to count exemplars. This is functionally correct but unnecessarily quadratic. Simplify:

```go
} else {
    exemplarCount := 0
    for _, t := range allTapes {
        if t.Exemplar {
            exemplarCount++
        }
    }
    if exemplarCount > 0 {
        logger.Printf("WARNING: %d exemplar tape(s) found but synthesis is disabled (use --synthesize to enable)", exemplarCount)
    }
}
```

Non-blocking.

server.go
```diff
@@ -391,24 +434,112 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		} else if params == nil {
+			// Pattern has no parameters -- check exact regex match.
+			if !ppc.re.MatchString(r.URL.Path) {
+				continue
+			}
+			params = map[string]string{}
+		}
+
+		// Score the path match using the PathPatternCriterion's score logic.
+		// We create a temporary tape with the URL set to the request path so
```
tibtof: **Stale comment**: the comment says *"We create a temporary tape with the URL set to the request path so that the criterion's Score method can compare"* but no temporary tape is actually created. The code directly uses `pathScore := 3` on the next line. Remove or rewrite this comment to match the actual implementation.

templating.go
```diff
@@ -840,3 +840,217 @@ func readRequestBody(r *http.Request) []byte {
...
+		rawExpr := exprs[0].raw
+		exprBody, coercion, hasCoercion := parseCoercion(rawExpr)
+
+		resolved, ok := resolveExpr(exprBody, ctx)
+		if !ok {
+			if strict {
+				*errOut = fmt.Errorf("httptape: unresolvable template expression: {{%s}}", rawExpr)
+				return s
+			}
+			if hasCoercion {
```
tibtof: **Redundant branch**: both arms of this conditional return `""`. The `if hasCoercion` check is unnecessary. Simplify to just `return ""`.

```go
if !ok {
    if strict {
        *errOut = fmt.Errorf(...)
        return s
    }
    return ""
}
```

cmd/httptape/main_test.go
```diff
@@ -879,6 +879,88 @@ func TestMigrateFixtures_AlreadyMigratedTape(t *testing.T) {
...
+			"body": null,
+			"body_hash": ""
+		},
+		"response": {
+			"status_code": 200,
+			"headers": {},
+			"body": null
+		}
+	}`
+	_ = store
```
tibtof: **Test cleanup**: `store` is created but unused (suppressed with `_ = store`). Either use it to save the tape (instead of writing raw JSON with `os.WriteFile`), or remove its creation entirely. Writing raw JSON directly is fine for this test (testing malformed fixtures), but then the `store` variable is noise.

tibtof (commented):
LGTM -- APPROVE (submitted as comment because self-approval is blocked by GitHub).

**What was reviewed**: 4 production files (`tape.go`, `server.go`, `templating.go`, `cmd/httptape/main.go`), 5 test files, 5 docs/CHANGELOG files. ~1876 lines total.

**Build/test verification**: `go build ./...`, `go vet ./...`, and `go test -race -count=1 ./...` all pass. No new external dependencies. gofmt issues are pre-existing (not in PR files).

**ADR fidelity**: implementation matches ADR-43 faithfully on all four design checkpoints:
- Exemplar shape: `Exemplar bool` + `URLPattern string` with mutual exclusion validated.
- `WithSynthesis()`: opt-in, disabled by default, exemplars inert without it.
- Body resolution: Content-Type-aware dispatch (JSON with coercion, text via byte substitution, binary skipped).
- Two-phase match: exact tapes first, exemplar fallback second with multi-criteria scoring.

**Dev decision validation**:
1. `ExtractParams` + `re.MatchString` instead of `MatchAndCapture` -- **justified**. `MatchAndCapture` does not exist on `PathPatternCriterion`. The adaptation using `ExtractParams` + direct regex access is functionally equivalent.
2. Separate `walkJSONWithCoercion` instead of modifying `walkJSON` -- **justified**. The existing `walkJSON` mutates maps/slices in-place, which would corrupt exemplar tape data across concurrent requests. The new function creates fresh copies. Correct design choice.
3. Flat `pathScore = 3` plus regex specificity -- **acceptable with caveat**. Patterns at different depths are disambiguated by regex anchoring (the test case works). Same-depth literal-vs-parameterized patterns (e.g., `/users/me` vs `/users/:id`) tie and fall back to declaration order. Flagged as a non-blocking follow-up suggestion.

**Test coverage**: comprehensive. Validation (7 table-driven cases), synthesis mode (8 server tests), JSON coercion (12 cases), exemplar body resolution (5 dispatch tests), integration (2 end-to-end tests), CLI (3 tests), JSON round-trip (3 marshal/unmarshal tests).

**Non-blocking comments posted** (7 total): specificity tie-breaking suggestion, silent skip of invalid patterns, dead code in validation, O(n^2) counting, stale comment, redundant branch, unused test variable. All are code-quality nits -- none affect correctness.

PR #239, opened 2026-04-27
feat(#208): WithCacheLookupDisabled() option (replaces neverMatcher workaround)

### LINKED ISSUE #208: feat(caching): expose disable-cache-lookup as WithCacheLookupDisabled() (replace neverMatcher workaround)
## Context

In v0.13.1 (PR #207, ADR-45), \`Proxy\` was refactored to compose \`CachingTransport\` internally. The dev encountered a constraint: ADR-45's original \"L1 pre-check then delegate to CachingTransport\" approach broke a health integration test, because CachingTransport would serve from its store once a tape was recorded, preventing upstream from being consulted.

The dev worked around this with an unexported \`neverMatcher\`:

\`\`\`go
type neverMatcher struct{}
func (neverMatcher) Match(_ *http.Request, _ []Tape) (Tape, bool) { return Tape{}, false }
\`\`\`

This forces CachingTransport to always miss, so it acts as a record-only / single-flight / SSE-tee primitive at the Proxy layer while Proxy itself owns the hit-path lookup against L1.

The reviewer accepted this as pragmatic but recommended exposing the capability as a first-class API.

## Why this matters

Other Go embedders may want the same shape — use CachingTransport's record-on-miss + single-flight + SSE-tee + sanitization plumbing without its hit-path replay. The \`neverMatcher\` workaround is internal-only; external embedders can't easily replicate it.

## Proposed API

\`\`\`go
// WithCacheLookupDisabled disables the cache hit path entirely. Every
// request is treated as a miss: forwarded to upstream, recorded via the
// sanitization pipeline (if configured), and returned. Single-flight
// dedup, SSE tee, and sanitization remain active.
//
// Useful when the embedder owns its own hit-path logic (e.g., proxy.go
// uses an L1 store consulted before this transport runs) and wants
// CachingTransport's other cross-cutting concerns without the cache
// lookup it would otherwise perform.
func WithCacheLookupDisabled() CachingOption
\`\`\`

Internally this is equivalent to setting the matcher to a never-matches matcher (which is exactly what \`proxy.go\`'s \`neverMatcher\` does).

## Acceptance criteria

1. New \`WithCacheLookupDisabled()\` \`CachingOption\` in \`caching_transport.go\`.
2. Implementation replaces the matcher with an internal never-matches sentinel (or sets a flag that short-circuits the hit-path).
3. \`proxy.go\` uses \`WithCacheLookupDisabled()\` in place of the local \`neverMatcher\` type. The \`neverMatcher\` type is removed from \`proxy.go\` (no longer needed).
4. New test in \`caching_transport_test.go\` verifying that with \`WithCacheLookupDisabled()\`, even a request that would match an existing tape is forwarded to upstream (cache miss path runs).
5. Existing tests stay green (\`go test -race ./...\`).
6. Doc updates: \`docs/caching-transport.md\` documents the new option; \`docs/api-reference.md\` lists it.
7. CHANGELOG entry under v0.13.x or v0.14.0 (whichever ships first).

## Bonus cleanup (while in this code)

The reviewer also flagged a swallowed error in \`proxy.go\` around line 152 in the \`l1RecordingTransport.RoundTrip\` flow:

\`\`\`go
reqBody, _ = io.ReadAll(body)
\`\`\`

The \`io.ReadAll\` error is silently dropped. If the read fails, recording silently uses an empty body. Either:
- Add a comment explaining why the error is intentionally ignored (recording is best-effort, can't propagate via \`http.RoundTripper\` semantics)
- Or invoke the \`onError\` callback with a wrapped error
- Architect's call.

## Out of scope

- Don't refactor the broader \`l1RecordingTransport\` semantics — just the option exposure + swallowed-error decision.
- Don't add other CachingTransport configuration knobs.

## Dependencies

None. Independent feature.

tibtof:
Closes #208

## Summary

- Add `WithCacheLookupDisabled()` CachingOption that disables the cache hit path entirely. When set, `RoundTrip` skips `Store.List` and `Matcher.Match`, jumping straight to the miss path. Single-flight dedup, SSE tee, sanitization, and stale fallback remain active.
- Delete the unexported `neverMatcher` type from `proxy.go`. Proxy now uses `WithCacheLookupDisabled()` instead of `WithCacheMatcher(neverMatcher{})`.
- Fix swallowed `io.ReadAll` error in `l1RecordingTransport.RoundTrip` (proxy.go line 160) -- now invokes `onError` consistently with the response body read a few lines later.
- Update `docs/caching-transport.md` and `docs/api-reference.md` with new option documentation.
- Add CHANGELOG entry under `[Unreleased]`.

## Test plan

- [x] `TestCachingTransport_LookupDisabledAlwaysMisses` -- pre-seeded cache is bypassed, upstream is contacted
- [x] `TestCachingTransport_LookupDisabledStillRecords` -- responses are still persisted to the store
- [x] `TestCachingTransport_LookupDisabledWithStaleFallback` -- stale fallback still works when lookup is disabled (key semantic difference vs neverMatcher)
- [x] `TestCachingTransport_LookupDisabledSkipsStoreList` -- spy store confirms `Store.List` is NOT called (validates perf benefit)
- [x] All existing tests pass: `go test -race -count=1 ./...`
- [x] `go build ./...` and `go vet ./...` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
+- **`WithCacheLookupDisabled()`**: CachingOption that disables the cache
+  hit path entirely. Every request is forwarded to upstream and recorded.
+  Single-flight dedup, SSE tee, sanitization, and stale fallback remain
+  active. Replaces the internal `neverMatcher` workaround used by Proxy. (#208)
+
 - **Template helpers**: `{{now}}`, `{{uuid}}`, `{{randomHex}}`, `{{randomInt}}`,
   `{{counter}}`, and `{{faker.*}}` template expressions in response bodies and
   headers. Helpers support keyword arguments (e.g., `{{now format=unix}}`). (#196)
--- a/caching_transport.go
+++ b/caching_transport.go
@@ -31,9 +31,10 @@ type CachingTransport struct {
 	onError     func(error)
 	cacheFilter func(*http.Response) bool
 
-	singleFlight bool
-	maxBodySize  int
-	sseRecording bool
+	singleFlight   bool
+	lookupDisabled bool
+	maxBodySize    int
+	sseRecording   bool
 
 	// stale-fallback (#164)
 	staleFallback   bool
@@ -143,6 +144,25 @@ func WithCacheSSERecording(enabled bool) CachingOption {
 	}
 }
 
+// WithCacheLookupDisabled disables the cache hit path entirely.
+// Every request is treated as a miss: forwarded to upstream, recorded
+// via the sanitization pipeline (if configured), and returned.
+// Single-flight dedup, SSE tee, and sanitization remain active.
+//
+// The configured Matcher is still used by stale fallback
+// (WithCacheUpstreamDownFallback), so the two options compose:
+// disable the hit path but still serve stale tapes when upstream is down.
+//
+// Useful when the embedder owns its own hit-path logic (e.g., Proxy
+// uses an L1 store consulted before this transport runs) and wants
+// CachingTransport's other cross-cutting concerns without the cache
+// lookup it would otherwise perform.
+func WithCacheLookupDisabled() CachingOption {
+	return func(ct *CachingTransport) {
+		ct.lookupDisabled = true
+	}
+}
+
 // WithCacheUpstreamDownFallback enables stale-response fallback when
 // upstream is unreachable or returns a transport error on a cache miss.
 // When enabled, CachingTransport searches the store for the best-matching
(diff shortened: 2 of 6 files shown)


tibtof (commented):
**Reviewer Agent: APPROVE**

LGTM. Clean implementation that matches the architect's design exactly.

**Verified:**

- `lookupDisabled` boolean field + `WithCacheLookupDisabled()` option function with thorough godoc
- `RoundTrip` guard: entire `Store.List` + `Matcher.Match` block skipped (not just short-circuited), avoiding unnecessary I/O
- `serveStaleFallback` is independent of `lookupDisabled` — uses `ct.matcher.Match` directly, so the two options compose correctly
- `neverMatcher` type and all references removed from `proxy.go`; `NewProxy` now uses `WithCacheLookupDisabled()` with updated comments
- Bonus fix: swallowed `io.ReadAll` error in `l1RecordingTransport.RoundTrip` now reports via `onErrorSafe`, consistent with the adjacent response body error handling pattern
- Four well-structured tests: always-misses, still-records, stale-fallback composition, spy-store skips-List
- `spyStore` correctly implements `Store` via `*MemoryStore` embedding with atomic `List` counter
- Docs (`caching-transport.md`, `api-reference.md`) and CHANGELOG updated
- `go build`, `go vet`, `go test -race`, `gofmt` all clean
- No new dependencies, no architecture violations

PR #241, opened 2026-04-27
docs(#197): ADR-47 inbound TLS listener design

### LINKED ISSUE #197: feat(server): inbound TLS listener (self-signed cert or user-provided)
## Problem statement

httptape currently listens on plain HTTP only. Outbound TLS (httptape to upstream) is well-supported via `--tls-ca`, `--tls-cert`, `--tls-key`, `--tls-insecure`. But for test scenarios where the client SDK under test hardcodes HTTPS, or where test code branches on URL scheme, inbound plaintext breaks the test flow.

## User story

As a Go developer testing code that hardcodes HTTPS endpoints (e.g., LLM SDK clients that default to `https://api.openai.com`), I want httptape to serve HTTPS natively so that I can redirect my client to the mock without scheme-mismatch workarounds.

## Context

Today's workaround is "put a reverse proxy in front" — works in CI, but is annoying for embedded Go tests and quick local runs. For the LLM agent testing use case specifically, most SDK clients default to HTTPS and some do not allow overriding the scheme separately from the host.

### CLI flags

```bash
# User-provided cert + key:
httptape serve --fixtures ./fixtures \
  --tls-listener-cert ./cert.pem \
  --tls-listener-key ./key.pem

# Auto-generated self-signed cert at startup:
httptape serve --fixtures ./fixtures --tls-listener-auto
```

`--tls-listener-cert` + `--tls-listener-key` are mutually exclusive with `--tls-listener-auto`.

### ServerOptions

- `WithTLSListenerCert(certFile, keyFile string)` — user-provided cert.
- `WithTLSListenerAutoCert()` — generate self-signed at startup, log fingerprint.

### Auto-cert behavior

- Generates an ephemeral self-signed cert using `crypto/tls` + `crypto/x509` + `crypto/rsa` (all stdlib).
- Includes sensible SANs: `localhost`, `127.0.0.1`, configurable via `--tls-listener-san` flag.
- On startup, logs cert fingerprint (SHA-256) and PEM-encoded public cert so tests can pin/trust it programmatically.

## Acceptance criteria

- [ ] `--tls-listener-cert` + `--tls-listener-key` flags, mutually exclusive with `--tls-listener-auto`.
- [ ] `WithTLSListenerCert` and `WithTLSListenerAutoCert` ServerOptions.
- [ ] Self-signed cert generation uses `crypto/tls` + `crypto/x509` + `crypto/rsa` (stdlib). Includes sensible SANs (`localhost`, `127.0.0.1`, configurable via `--tls-listener-san` flag).
- [ ] On startup, log cert fingerprint and PEM-encoded public cert so tests can pin/trust it programmatically.
- [ ] Integration test: start server with auto-cert, hit it with a TLS-verifying client that trusts the generated cert, verify end-to-end.
- [ ] Docs: update `docs/tls.md` with inbound section; update `docs/cli.md`.

1. **Certificate expiry**: Generated cert validity period? PM recommendation: 24 hours (ephemeral; server is expected to be short-lived in tests).
2. **Listening port behavior**: Default stays 8081 for HTTP; if `--tls-listener-*` is set, does port change to 8443? Or does httptape listen on both 8081 (HTTP) and 8443 (HTTPS)? PM leans: switch to 8443 when TLS is enabled, because dual-listen is a can of worms.
3. **Key algorithm**: RSA 2048 is fine for test certs? Or ECDSA P-256 for faster generation? Architect's call.

## Dependencies

None.

## Out of scope

- Inbound mTLS (client cert verification) — follow-up issue.
- ALPN / HTTP/2 over TLS tuning — stdlib defaults are fine.
- Certificate rotation at runtime — start server, serve, die. No hot-reload.

tibtof:
## Summary

ADR-47 for #197: design for inbound TLS listener support. Key decisions:

- `GenerateSelfSignedCert(hosts ...string) (*SelfSignedCert, error)` library function in `tls.go`
- ECDSA P-256, 24h validity, in-memory only (never touches disk)
- CLI flags `--tls-listener-cert`, `--tls-listener-key`, `--tls-listener-auto`, `--tls-listener-san` on all three listener commands (serve, record, proxy)
- No `ServerOption` -- `Server` stays a pure `http.Handler`; TLS is the caller's responsibility
- Port stays 8081 (no magic switch to 8443)
- Inbound mTLS deferred to follow-up issue

## Test plan

- [ ] ADR is internally consistent
- [ ] No contradiction with CLAUDE.md locked decisions
- [ ] No contradiction with existing ADRs in decisions.md

## Related

Closes #197 (design phase)

### DIFF
--- a/decisions.md
+++ b/decisions.md
@@ -16171,3 +16171,386 @@ conforming to it.
   changes needed.
 - **No impact on `NewRecorder`, `NewCachingTransport`, `NewMemoryStore`,
   `NewHealthMonitor`** — their signatures remain unchanged.
+
+---
+
+### ADR-47: Inbound TLS listener (self-signed cert or user-provided)
+
+**Date**: 2026-04-23
+**Issue**: #197
+**Status**: Accepted
+
+#### Context
+
+httptape currently listens on plain HTTP only. Outbound TLS (httptape to
+upstream) is well-supported via `BuildTLSConfig` and the `--tls-cert`,
+`--tls-key`, `--tls-ca`, `--tls-insecure` CLI flags. But for test scenarios
+where the client SDK under test hardcodes HTTPS (e.g., LLM SDK clients that
+default to `https://api.openai.com`), inbound plaintext breaks the test flow.
+
+The current workaround is "put a reverse proxy in front" which works in CI
+but is annoying for embedded Go tests and quick local runs.
+
+This ADR covers two capabilities:
+
+1. A library function that generates an ephemeral self-signed TLS certificate
+   (zero-config TLS for local dev and tests).
+2. CLI flags for all three listener commands (`serve`, `record`, `proxy`) to
+   enable TLS on the inbound listener, either with a user-provided cert/key
+   pair or with an auto-generated self-signed cert.
+
+##### Key architectural constraint
+
+`Server` is an `http.Handler`. It does not own a listener. Per CLAUDE.md
+layer rules, core types and services have zero I/O. The TLS listener is
+therefore the responsibility of the **caller** (the CLI, or a Go embedder
+using `net/http/httptest.NewUnstartedServer` + `StartTLS`). The library's
+role is limited to providing a helper function that generates a `tls.Config`
+from a self-signed cert.
+
+This design keeps the hexagonal boundary clean: the library produces a
+`*tls.Config`, the CLI calls `httpServer.ServeTLS()` or
+`httpServer.ListenAndServeTLS()`, and Go embedders use `httptest.NewTLSServer`
+or equivalent. No `ServerOption` is needed.
+
+#### Decision
+
+##### No new ServerOption types
+
+The PM spec proposed `WithTLSListenerCert` and `WithTLSListenerAutoCert`
+ServerOptions. This ADR rejects that approach. `Server` is an `http.Handler`.
+Adding TLS listener state to it would violate the hexagonal boundary (core
+service owns I/O) and couple the handler to a specific listener strategy.
+
+Instead, TLS is configured at the listener layer:
+
+- **CLI**: flags on the CLI command, calling `ServeTLS` or
+  `ListenAndServeTLS` on the `http.Server`.
+- **Go library embedders**: use `GenerateSelfSignedCert` to get a
+  `tls.Certificate`, then configure their own `tls.Config` and pass it to
+  `httptest.NewUnstartedServer` or `net/http.Server.TLSConfig`.
+
+This matches how Go's standard library works: `http.Handler` knows nothing
+about TLS; `http.Server` owns the listener.
+
+##### New library function: `GenerateSelfSignedCert`
+
+```go
+// SelfSignedCert holds an ephemeral self-signed TLS certificate and its
+// metadata. The certificate is generated in-memory and never touches disk.
+type SelfSignedCert struct {
+    // TLSCertificate is the parsed TLS certificate suitable for use in
+    // tls.Config.Certificates.
+    TLSCertificate tls.Certificate
+
+    // CertPEM is the PEM-encoded X.509 certificate. Useful for writing to
+    // a file, logging, or configuring client trust.
+    CertPEM []byte
+
+    // Fingerprint is the SHA-256 fingerprint of the DER-encoded certificate,
+    // hex-encoded with colon separators (e.g., "AB:CD:EF:..."). Useful for
+    // pinning and log identification.
+    Fingerprint string
+}
+
+// GenerateSelfSignedCert creates an ephemeral self-signed TLS certificate
+// suitable for inbound HTTPS listeners in tests and local development.
+//
+// The certificate is valid for the given hosts (DNS names and/or IP addresses).
+// If hosts is empty, the certificate covers "localhost" and "127.0.0.1".
+//
+// The generated certificate:
+//   - Uses ECDSA P-256 for fast generation and small key size.
+//   - Is valid for 24 hours from the current time.
+//   - Includes both DNS SANs and IP SANs based on the provided hosts.
+//   - Is generated in-memory and never touches disk.
+//
+// Returns an error if key generation or certificate creation fails.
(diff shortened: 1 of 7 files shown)


decisions.md
```diff
@@ -16171,3 +16171,386 @@ conforming to it.
...
+The TLS listener flags apply to all three commands that start an HTTP
+listener: `serve`, `record`, and `proxy`. The PM spec only mentions `serve`,
+but `record` and `proxy` also listen on a port and face the same scheme-
+mismatch problem.
+
+```
+--tls-listener-cert <path>   Path to PEM certificate for inbound TLS
+--tls-listener-key  <path>   Path to PEM private key for inbound TLS
+--tls-listener-auto          Generate a self-signed cert at startup
+--tls-listener-san  <hosts>  Comma-separated SANs for auto-cert (default: localhost,127.0.0.1)
```
tibtof, line 16324: **Inconsistency**: the Default SANs section states that `GenerateSelfSignedCert()` with no args includes `::1` as an IP SAN. But this CLI flag default is `localhost,127.0.0.1` (no `::1`), and the auto-cert sequence on line 16431 calls `GenerateSelfSignedCert("localhost", "127.0.0.1")` -- which explicitly provides hosts, so the function's no-args default (that includes `::1`) is bypassed.

This means CLI users with `--tls-listener-auto` (no `--tls-listener-san`) will get a cert that does NOT cover IPv6 loopback, even though the library function's default does.

**Suggestion**: Either (a) change the `--tls-listener-san` default to `localhost,127.0.0.1,::1` and update the sequence example, or (b) specify that the CLI calls `GenerateSelfSignedCert()` with no args when `--tls-listener-san` is not provided (delegating to the function default), and only passes explicit hosts when `--tls-listener-san` IS provided. Option (b) is cleaner -- it avoids duplicating the default in two places.

decisions.md
```diff
@@ -16171,3 +16171,386 @@ conforming to it.
...
+    // hex-encoded with colon separators (e.g., "AB:CD:EF:..."). Useful for
+    // pinning and log identification.
+    Fingerprint string
+}
+
+// GenerateSelfSignedCert creates an ephemeral self-signed TLS certificate
+// suitable for inbound HTTPS listeners in tests and local development.
+//
+// The certificate is valid for the given hosts (DNS names and/or IP addresses).
+// If hosts is empty, the certificate covers "localhost" and "127.0.0.1".
```
tibtof, line 16260: **Inconsistency (same root cause as the CLI flag default)**: This godoc says defaults to `localhost` and `127.0.0.1`, but the "Default SANs" section below lists three defaults: `localhost`, `127.0.0.1`, AND `::1`. One of them is wrong.

**Suggestion**: Decide whether `::1` is in the no-args default or not, then make the godoc, the Default SANs section, and the CLI flag default all agree. Including `::1` in the no-args default is the more useful choice (IPv6 loopback is common in modern stacks).

tibtof (commented):
## Review: REQUEST_CHANGES

(Note: GitHub does not allow requesting changes on your own PR, so this is posted as a comment review. The verdict is REQUEST_CHANGES.)

### ADR-47 (inbound TLS listener design)

The ADR is well-structured, thorough, and architecturally sound. The decision to keep `Server` as a pure `http.Handler` and place TLS at the caller/CLI layer is correct and consistent with CLAUDE.md hexagonal boundaries. The cert algorithm choice (ECDSA P-256), validity window (24h), mutual exclusion rules, and test strategy are all well-reasoned.

**One inconsistency requires resolution before this ADR can guide implementation:**

The `::1` (IPv6 loopback) default is stated inconsistently across three locations in the ADR:
1. The `GenerateSelfSignedCert` godoc says defaults to `localhost` and `127.0.0.1` (2 entries, no `::1`)
2. The "Default SANs" section says no-args defaults include `localhost`, `127.0.0.1`, AND `::1` (3 entries)
3. The `--tls-listener-san` flag default is `localhost,127.0.0.1` (2 entries, no `::1`)
4. The auto-cert CLI sequence calls `GenerateSelfSignedCert("localhost", "127.0.0.1")` which bypasses the function default entirely

The dev agent will need a single source of truth. See inline comments for the suggested fix.

### WithCacheLookupDisabled (#208)

The code change is clean, well-tested (4 tests covering bypass, recording, stale-fallback composition, and Store.List skip verification), and correctly replaces the `neverMatcher` workaround. The fix to the swallowed error in `l1RecordingTransport.RoundTrip` (proxy.go) is a good catch. No issues found with the code portion.

### Checklist verdicts

- **Layer rules**: PASS -- `Server` stays as `http.Handler`, `tls.go` is a Helper, no listener concerns added to services
- **Locked decisions**: PASS -- stdlib only, no init/state, no panic in hot paths
- **ADR style**: PASS -- matches ADR-44/45/46 structure (Date, Issue, Status, Context, Decision, Types, Functions, File layout, Sequence, Error cases, Test strategy, Alternatives, Consequences)
- **Backward compatibility**: PASS -- purely additive, no breaking changes
- **CLI flag mutual exclusion**: PASS -- clearly documented
- **Cert algorithm**: PASS -- ECDSA P-256 justified by speed, size, and consistency with existing test helpers
- **Fingerprint format**: PASS -- SHA-256, colon-separated hex, clearly documented
- **Test scope**: PASS -- 7 unit tests + 1 integration test + 3 CLI validation tests specified
- **Cross-cutting (record/proxy flags)**: PASS -- all three listener commands get the same four flags
- **Default SANs consistency**: FAIL -- `::1` included/excluded inconsistently (see inline comments)

tibtof:
Closing without merge — this ADR isn't warranted.

After re-evaluation, the substance of ADR-47 ("TLS lives in CLI/listener layer, not as ServerOption") is just *applying* CLAUDE.md's existing hexagonal layer rule (Server is `http.Handler`, owns no I/O). It's not a new architectural decision.

Tightened the project rule: ADRs are reserved for changes to locked decisions, public API contract breaks, or genuine multi-option architectural tradeoffs — NOT feature designs that follow established patterns.

The full design is being re-posted as a status comment on issue #197 so the dev has a single source of truth. Implementation proceeds without an ADR.

Refs #197.

PR #248, opened 2026-04-27
feat(#240): record response elapsed time + opt-in replay delay

### LINKED ISSUE #240: feat: record response elapsed time + opt-in replay delay
## Problem

Mock servers replay responses instantly. Tests for retry logic, timeouts, latency-sensitive client code, and race conditions can't easily reproduce real-world timing because the recorded fixtures don't carry timing information.

SSE already has per-event timing (`SSEEvent.OffsetMS` + `SSETimingMode` with `Recorded`/`Accelerated(factor)`/`Instant`). This issue completes the picture for non-SSE responses.

## User story

As a Go developer writing latency-sensitive tests (retry logic, timeout handling, circuit breakers), I want recorded HTTP fixtures to carry response elapsed time and replay it on demand, so that my test mock reproduces real-world timing behavior without manual sleep calls.

### Recording: always-on, no flag

Every recording captures `RecordedResp.ElapsedMS int64` (total elapsed from request issue to full response received). Instrument the recording paths via a single shared helper so behavior is consistent:
- `Recorder.RoundTrip` (the `http.RoundTripper` wrapper)
- `CachingTransport.roundTripUpstream` (cache miss path)
- `Proxy`'s upstream call (transitively via CachingTransport, so likely just verify)

### Replay: opt-in, default off

Mirrors the existing SSE timing model. New `ResponseTimingMode` interface with three constructors:
- `Instant` — no delay (default behavior; backwards-compatible with pre-feature behavior)
- `Accelerated(factor float64)` — scale recorded delay by factor (e.g., 0.1 = 10% of recorded)
- `Recorded` — sleep for the full recorded `ElapsedMS`

New options:
- `WithReplayTiming(mode ResponseTimingMode) ServerOption` on the Server replay path
- Equivalent on `CachingTransport` replay path (`WithCacheReplayTiming` or similar — architect to pick the exact name)

### Schema: additive

- New field: `RecordedResp.ElapsedMS int64`
- JSON serialization: `"elapsed_ms": int64` (or omit when zero)
- Backward-compatible: pre-feature fixtures (no `ElapsedMS` field, or zero value) replay with **no delay regardless of mode** (zero = unknown, conservative behavior)

### Where replay applies

- Server replay path (`Server.ServeHTTP` body write)
- CachingTransport replay path (cache hit returns)
- Proxy uses CachingTransport, so transitively covered

## Acceptance criteria

- [ ] `RecordedResp.ElapsedMS int64` field added to the core type
- [ ] JSON encoding handles the new field (the alias structs in `RecordedResp.MarshalJSON` and `RecordedResp.UnmarshalJSON` in `tape.go` must include the field)
- [ ] Recording helper instruments all recording paths — single function, consistent behavior. Note: there are more than three places where `RecordedResp` is constructed (see "Recording paths inventory" below). The shared helper must cover all of them.
- [ ] `ResponseTimingMode` interface + three constructors (`Instant`, `Accelerated(factor)`, `Recorded`), mirroring `SSETimingMode` shape
- [ ] `WithReplayTiming(mode)` option on `Server`
- [ ] Equivalent option on `CachingTransport` (architect to name)
- [ ] **Recording is always on** — no opt-out, no flag, no option
- [ ] **Replay delay is opt-in** — default mode is `Instant` (no delay)
- [ ] Pre-feature fixtures (no `ElapsedMS` or zero value) replay with no delay regardless of mode
- [ ] Tests:
  - Recording captures elapsed time within reasonable tolerance
  - `Recorded` mode delays approximately the recorded duration
  - `Accelerated(0.1)` mode delays ~10% of recorded
  - `Instant` mode skips delay entirely (default)
  - Pre-feature fixtures don't delay even in `Recorded` mode
  - Race-clean: all timing-sensitive tests run under `-race` and are not flaky
  - Clock/sleep injection for tests: tests must NOT actually sleep for real durations in CI. Abstract `time.Now`/`time.Sleep` behind an injectable dependency (function field, interface, or similar) so tests use a fake clock.
- [ ] Sanitization pipeline does not redact `ElapsedMS` (it's not sensitive but should be respected)
- [ ] CHANGELOG under `[Unreleased]` describing additive schema field + new option
- [ ] `docs/recording.md` and `docs/replay.md` updated; new section in `docs/replay.md` for timing modes if not already there for SSE

## Recording paths inventory (architect: verify completeness)

The codebase currently constructs `RecordedResp` in these locations — the shared elapsed-time helper must cover all of them:

1. **`Recorder.RoundTrip`** (`recorder.go` ~line 340) — standard non-SSE recording
2. **`Recorder.roundTripSSE`** (`recorder.go` ~line 412) — SSE recording in `onDone` callback
3. **`CachingTransport.roundTripUpstream`** (`caching_transport.go` ~line 471) — cache miss, non-SSE
4. **`CachingTransport.roundTripSSE`** (`caching_transport.go` ~line 540) — cache miss, SSE, in `onDone` callback
5. **`l1RecordingTransport.RoundTrip`** (`proxy.go` ~line 189) — L1 raw tape save, non-SSE
6. **`l1RecordingTransport.roundTripSSE`** (`proxy.go` ~line 247) — L1 raw tape save, SSE, in `onDone` callback

Paths 5 and 6 are inside the Proxy's L1 recording wrapper. The issue says "Proxy is transitively covered via CachingTransport" — this is true for L2, but L1's `l1RecordingTransport` builds its own `RecordedResp` independently. The architect must decide whether L1 tapes also get `ElapsedMS` (they should, for consistency).

### OQ-1: JSON encoding — omit when zero or always emit?

Should `elapsed_ms` be omitted from JSON when zero (`omitempty` / conditional in MarshalJSON alias), or always emitted?

**PM recommendation**: Omit when zero. Rationale: matches Go zero-value convention, keeps pre-feature fixture files byte-identical after re-serialization, and is consistent with other optional fields in `RecordedResp` (e.g., `Truncated`, `OriginalBodySize`, `SSEEvents` all use `omitempty`).

### OQ-2: Server option name

Is `WithReplayTiming(mode ResponseTimingMode)` the right name? Note the existing `WithSSETiming(mode SSETimingMode)` in `server.go` — should these be parallel (`WithReplayTiming` for HTTP, `WithSSETiming` for SSE)?

**PM recommendation**: `WithReplayTiming` is good. The SSE timing mode controls inter-event gaps; the response timing mode controls the initial response delay. These are different enough to warrant distinct names.

### OQ-3: CachingTransport option name

What should the CachingTransport equivalent be? Existing cache options follow `WithCache*` pattern: `WithCacheMatcher`, `WithCacheSanitizer`, `WithCacheFilter`, `WithCacheSSERecording`, etc.

**PM recommendation**: `WithCacheReplayTiming(mode ResponseTimingMode)` for consistency with the `WithCache*` prefix convention.

### OQ-4: Where to apply the delay

- **Server**: before writing the response body? After writing headers? Before the entire `WriteHeader` call?
- **CachingTransport**: before returning the `*http.Response` from `tapeToResponse`?

The user-visible contract should be: "the client perceives the response after `ElapsedMS` from when they made the request." Note: Server already has a delay mechanism (the `effectiveDelay` block at ~line 359 in `server.go`) that sleeps before writing the response. The new replay timing could compose with or replace this.

**PM recommendation**: The architect should clarify the interaction between `WithDelay` (fixed delay), per-fixture `metadata.delay`, and `WithReplayTiming` (recorded delay). All three could be active simultaneously. Recommend: replay timing is additive (applied on top of fixed/per-fixture delay), not a replacement. Document the ordering.

### OQ-5: Sleep mechanism

`time.Sleep` (simple, not cancellable) vs `time.NewTimer` + context cancellation (cancellable, respects client disconnect)?

**PM recommendation**: Timer + context cancellation. The existing `effectiveDelay` in `server.go` already uses this pattern (timer + `select` on `timer.C` and `r.Context().Done()`). The SSE replay in `replaySSEEvents` also uses this pattern. Follow precedent.

### OQ-6: Should `Accelerated(factor)` allow factor > 1?

Factor > 1 would *slow down* the replay (e.g., factor 2.0 = double the recorded delay). Is this useful or should it be clamped to (0, 1]?

**PM recommendation**: Allow factor > 1. The SSE precedent (`SSETimingAccelerated`) only validates `factor > 0` and does not cap at 1.0. A factor > 1 means "slower than recorded," which can be useful for stress-testing timeout paths. The naming "Accelerated" is slightly misleading for factor > 1, but changing SSE's convention is out of scope.

### OQ-7: Naming discrepancy — `Recorded` vs `Realtime`

The issue specifies `Recorded` as the constructor for "sleep the full `ElapsedMS`." However, the SSE precedent uses `SSETimingRealtime()` for the equivalent concept (replay with original timing). If the goal is to "mirror `SSETimingMode`," should the constructor be `ResponseTimingRealtime()` or `ResponseTimingRecorded()`?

**PM recommendation**: The semantics differ enough to justify different names. SSE's `Realtime` replays inter-event *gaps*; this feature replays a single total elapsed time. `Recorded` is clearer for users ("replay at the recorded speed"). The architect should make the final call, but document the naming rationale if it diverges from SSE.

### OQ-8: Clock injection strategy

Tests must not actually sleep for real durations. The architect must decide the injection mechanism:
- **Function fields** (e.g., `nowFunc func() time.Time`, `sleepFunc func(time.Duration)`) on Recorder/Server/CachingTransport — simplest, used in many Go projects
- **Clock interface** (e.g., `type Clock interface { Now() time.Time; NewTimer(d) *time.Timer }`) — more structured, harder to mock `*time.Timer`
- **Inject only into the timing mode** (the `ResponseTimingMode.delay()` method returns a `time.Duration`, and the sleep mechanism is injected where the delay is applied)

**PM recommendation**: The simplest approach that achieves non-flaky tests. Follow the pattern already in use for similar concerns (e.g., `randFloat func() float64` on Server for error-rate testing). Function fields are likely sufficient.

## Out of scope (defer to follow-ups)

- TTFB (time-to-first-byte) capture — additive later if a use case appears
- Per-route timing strategies (different modes for different routes)
- Jitter / randomization of replay delays
- Migration tool to backfill `ElapsedMS` on old fixtures (zero stays zero)
- Proxy-specific `WithProxyReplayTiming` option (Proxy's L1/L2 fallback path has its own `tapeToResponse` — if delay is wanted on fallback, that's a separate story)

## Notes for architect

- The `SSETimingMode` shape in `sse.go` is the precedent — match it as closely as makes sense.
- Decide whether `ResponseTimingMode` and `SSETimingMode` should share a parent interface or stay separate. Lean toward separate (different domains, different defaults).
- Single helper for recording elapsed time avoids drift between the recording paths. But note there are 6 recording sites, not 3 (see inventory above).
- The Server option might want to apply to *all* responses or be per-route. Default: applies to all responses; per-route can be a follow-up if requested.
- `RecordedResp.MarshalJSON` and `UnmarshalJSON` use explicit alias structs — the new field must be added to both aliases.

tibtof:
Closes #240

## Summary

- **`RecordedResp.ElapsedMS int64`** (`omitempty`): records total response elapsed time in milliseconds. Always-on recording, no opt-out. Both `MarshalJSON` and `UnmarshalJSON` alias structs updated.
- **`elapsedMS(start, nowFunc)` helper** in `tape.go`: used by all 6 recording sites.
- **`ResponseTimingMode` sealed interface** (mirrors `SSETimingMode` shape) with 3 constructors:
  - `ResponseTimingInstant()` -- no delay (default)
  - `ResponseTimingRecorded()` -- replay with recorded elapsed time
  - `ResponseTimingAccelerated(factor)` -- scale elapsed time by factor (>0 required)
- **`WithReplayTiming(mode)`** `ServerOption` on `Server`
- **`WithCacheReplayTiming(mode)`** `CachingOption` on `CachingTransport`
- **`defaultSleepFunc`**: timer + select on ctx.Done, same pattern as existing `effectiveDelay`

### Six recording sites (ElapsedMS populated)

1. `recorder.go` -- `Recorder.RoundTrip` non-SSE path
2. `recorder.go` -- `Recorder.roundTripSSE` SSE path
3. `caching_transport.go` -- `roundTripUpstream` non-SSE path
4. `caching_transport.go` -- `roundTripSSE` SSE path
5. `proxy.go` -- `l1RecordingTransport.RoundTrip` non-SSE path
6. `proxy.go` -- `l1RecordingTransport.roundTripSSE` SSE path

### Additive composition with WithDelay / metadata.delay

Replay timing runs AFTER the existing `effectiveDelay` block (step 9) and BEFORE writing the response (step 10). The total delay is their sum:

- `WithDelay(100ms)` + `ResponseTimingRecorded()` with `ElapsedMS=200` = 300ms total
- Documented in code comment at the apply site + godoc on `WithReplayTiming`

### Clock and sleep injection

- Each recording struct gains `nowFunc func() time.Time` (mirrors `randFloat` pattern)
- Server and CachingTransport gain `sleepFunc func(context.Context, time.Duration) error`
- All unexported `withNowFunc` / `withSleepFunc` test options prevent real sleeping

### Backward compatibility

- `elapsed_ms` uses `omitempty`: pre-feature fixtures remain byte-identical
- Default mode is `Instant` (no delay): pre-feature behavior preserved
- Pre-feature fixtures (`ElapsedMS == 0`) incur no delay regardless of mode

## Test plan

- [x] `RecordedResp` marshal/unmarshal round-trip with `ElapsedMS`
- [x] `ElapsedMS` omitted when zero (pre-feature compat)
- [x] `elapsedMS()` helper unit test
- [x] `ResponseTimingInstant` returns zero for all inputs
- [x] `ResponseTimingRecorded` returns recorded duration, zero for 0/negative
- [x] `ResponseTimingAccelerated` scales correctly, rejects factor <= 0
- [x] `WithReplayTiming(Recorded)` invokes sleepFunc with correct duration
- [x] `WithReplayTiming(Instant)` does not invoke sleepFunc
- [x] Pre-feature fixture (ElapsedMS=0) does not delay with `Recorded` mode
- [x] Additive composition: WithDelay + WithReplayTiming both apply
- [x] Context cancellation during replay timing aborts response
- [x] `defaultSleepFunc` zero/negative duration returns immediately
- [x] `defaultSleepFunc` cancelled context returns error
- [x] Recorder non-SSE records ElapsedMS via injected clock
- [x] Recorder SSE records ElapsedMS via injected clock
- [x] CachingTransport non-SSE records ElapsedMS via injected clock
- [x] CachingTransport SSE records ElapsedMS via injected clock
- [x] CachingTransport cache-hit with `WithCacheReplayTiming(Recorded)` sleeps
- [x] CachingTransport cache-hit pre-feature fixture does not sleep
- [x] l1RecordingTransport non-SSE records ElapsedMS
- [x] l1RecordingTransport SSE records ElapsedMS
- [x] Proxy records ElapsedMS in both L1 and L2
- [x] All tests race-clean (`go test -race`)
- [x] No real sleeping in tests (all use `sleepFunc` injection)
- [x] `go build ./...`, `go vet ./...`, `go test -race -count=1 ./...` all clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
+- **`RecordedResp.ElapsedMS`**: records total response time in milliseconds
+  for every HTTP interaction (non-SSE and SSE). Always-on, no opt-out.
+  Pre-feature fixtures (missing or zero `elapsed_ms`) are backward
+  compatible. (#240)
+
+- **`ResponseTimingMode`** sealed interface with three constructors for
+  opt-in replay delay based on recorded elapsed time:
+  - `ResponseTimingInstant()` -- no delay (default, preserves pre-feature behavior)
+  - `ResponseTimingRecorded()` -- replay with recorded elapsed time
+  - `ResponseTimingAccelerated(factor)` -- scale recorded elapsed time by factor
+
+- **`WithReplayTiming(mode)`**: ServerOption that applies replay timing
+  delay. Composes additively with `WithDelay` / `metadata.delay`.
+  Pre-feature fixtures incur no delay regardless of mode. (#240)
+
+- **`WithCacheReplayTiming(mode)`**: CachingOption that applies replay
+  timing delay on cache-hit responses. (#240)
+
 - **`WithCacheLookupDisabled()`**: CachingOption that disables the cache
   hit path entirely. Every request is forwarded to upstream and recorded.
   Single-flight dedup, SSE tee, sanitization, and stale fallback remain
--- a/caching_transport.go
+++ b/caching_transport.go
@@ -36,6 +36,11 @@ type CachingTransport struct {
 	maxBodySize    int
 	sseRecording   bool
 
+	// response timing
+	replayTiming ResponseTimingMode
+	nowFunc      func() time.Time
+	sleepFunc    func(context.Context, time.Duration) error
+
 	// stale-fallback (#164)
 	staleFallback   bool
 	upstreamTimeout time.Duration
@@ -188,6 +193,33 @@ func WithCacheUpstreamTimeout(d time.Duration) CachingOption {
 	}
 }
 
+// WithCacheReplayTiming sets the response timing mode for cache-hit
+// responses. Defaults to ResponseTimingInstant() (no delay, preserving
+// pre-feature behavior).
+//
+// Timing composition: WithCacheReplayTiming composes ADDITIVELY with any
+// delays the caller adds after receiving the response. Pre-feature
+// fixtures (ElapsedMS == 0) incur no replay timing delay regardless
+// of mode. The delay is applied in tapeToResponse before returning the
+// *http.Response, so the caller of RoundTrip perceives the delay.
+func WithCacheReplayTiming(mode ResponseTimingMode) CachingOption {
+	return func(ct *CachingTransport) {
+		ct.replayTiming = mode
+	}
+}
+
+// withCacheNowFunc overrides the clock for elapsed-time measurement.
+// Unexported -- only used in tests.
+func withCacheNowFunc(fn func() time.Time) CachingOption {
+	return func(ct *CachingTransport) { ct.nowFunc = fn }
+}
+
+// withCacheSleepFunc overrides the sleep function for testing.
+// Unexported -- only used in tests.
+func withCacheSleepFunc(fn func(context.Context, time.Duration) error) CachingOption {
+	return func(ct *CachingTransport) { ct.sleepFunc = fn }
+}
+
 // defaultCacheMaxBodySize is the default maximum request body size for cache
 // participation: 10 MiB.
 const defaultCacheMaxBodySize = 10 * 1024 * 1024
@@ -229,6 +261,9 @@ func NewCachingTransport(upstream http.RoundTripper, store Store, opts ...Cachin
 		singleFlight: true,
 		maxBodySize:  defaultCacheMaxBodySize,
 		sseRecording: true,
+		replayTiming: ResponseTimingInstant(),
+		nowFunc:      time.Now,
+		sleepFunc:    defaultSleepFunc,
 		cacheFilter: func(resp *http.Response) bool {
 			return resp.StatusCode >= 200 && resp.StatusCode < 300
 		},
@@ -290,7 +325,7 @@ func (ct *CachingTransport) RoundTrip(req *http.Request) (*http.Response, error)
 				if tape.Response.IsSSE() {
 					return ct.sseResponseFromTape(tape), nil
 				}
-				return ct.tapeToResponse(tape), nil
+				return ct.tapeToResponse(req.Context(), tape), nil
 			}
 		}
 	}
@@ -419,6 +454,7 @@ func (ct *CachingTransport) roundTripUpstream(req *http.Request, reqBody []byte,
 		req = req.WithContext(ctx)
 	}
 
+	startTime := ct.nowFunc()
 	resp, transportErr := ct.upstream.RoundTrip(req)
 	if transportErr != nil {
 		// Upstream error path.
@@ -439,7 +475,7 @@ func (ct *CachingTransport) roundTripUpstream(req *http.Request, reqBody []byte,
 
 	// SSE detection on miss path.
 	if ct.sseRecording && isSSEContentType(resp.Header.Get("Content-Type")) {
-		return ct.roundTripSSE(req, resp, reqBody)
+		return ct.roundTripSSE(req, resp, reqBody, startTime)
 	}
 
 	// Read response body into buffer.
@@ -472,6 +508,7 @@ func (ct *CachingTransport) roundTripUpstream(req *http.Request, reqBody []byte,
 		StatusCode: resp.StatusCode,
 		Headers:    resp.Header.Clone(),
 		Body:       respBody,
+		ElapsedMS:  elapsedMS(startTime, ct.nowFunc),
 	}
 
 	tape := NewTape(ct.route, recordedReq, recordedResp)
@@ -492,8 +529,8 @@ func (ct *CachingTransport) roundTripUpstream(req *http.Request, reqBody []byte,
 // while a background goroutine parses SSE events. When the stream
 // completes cleanly, the tape is persisted. If the stream is truncated
 // (client disconnect), the partial tape is discarded.
-func (ct *CachingTransport) roundTripSSE(req *http.Request, resp *http.Response, reqBody []byte) (*http.Response, error) {
-	startTime := time.Now()
+func (ct *CachingTransport) roundTripSSE(req *http.Request, resp *http.Response, reqBody []byte, startTime time.Time) (*http.Response, error) {
+	nowFunc := ct.nowFunc // capture for closure
 	respHeaders := resp.Header.Clone()
 
 	recordedReq := RecordedReq{
@@ -542,6 +579,7 @@ func (ct *CachingTransport) roundTripSSE(req *http.Request, resp *http.Response,
 			Headers:    respHeaders,
 			Body:       nil,
 			SSEEvents:  collectedEvents,
+			ElapsedMS:  elapsedMS(startTime, nowFunc),
 		}
 
 		tape := NewTape(ct.route, recordedReq, recordedResp)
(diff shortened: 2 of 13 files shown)


tibtof:
## Review verdict: APPROVE

(Posted as a comment because GitHub blocks self-review on author=reviewer.)

### Checks

stdlib only ✅ · backward compatibility (pre-feature fixtures replay with no delay regardless of mode, asserted by dedicated tests) ✅ · recording always-on (all 6 sites populate \`ElapsedMS\` unconditionally) ✅ · replay opt-in (\`Instant\` default in both \`NewServer\` and \`NewCachingTransport\`) ✅ · additive composition with \`WithDelay\`/\`metadata.delay\` documented in code comment + godoc ✅ · \`sleepFunc\` injection (no \`time.Sleep\` calls in test code; full suite under 6s) ✅ · \`startTime\` propagated cleanly from \`RoundTrip\` to \`roundTripSSE\` in all three sites ✅ · \`tapeToResponse\` context param uses \`context.Background()\` correctly on stale-fallback ✅ · JSON aliases include \`elapsed_ms,omitempty\` in both Marshal and Unmarshal ✅ · build/race/vet/fmt clean ✅ · CHANGELOG entry present ✅

### Non-blocking observation

**SSE asymmetry between Server and CachingTransport replay timing.**

In \`Server.ServeHTTP\`, replay timing (step 9b) runs **before** the SSE check (step 10), so SSE tapes receive the replay-timing delay plus SSE event timing. In \`CachingTransport\`, SSE cache hits go through \`sseResponseFromTape\` (not \`tapeToResponse\`), so they do **not** receive the replay-timing delay.

This matches the design spec ("apply the sleep in \`tapeToResponse\`" for CachingTransport), so it's intentional. Worth a follow-up issue if SSE replay timing matters for CachingTransport consumers — symmetric behavior would require applying the delay in \`sseResponseFromTape\` as well.

Not a blocker for this PR.

Issue #3, opened 2026-03-30
1.3 Mock server (http.Handler)


tibtof:
## Problem Statement

To replay recorded HTTP interactions, httptape needs a mock server that serves as an `http.Handler`. This allows developers to stand up a fake HTTP server backed by recorded fixtures, enabling deterministic integration tests without hitting real APIs.

## User Story

As a Go developer, I want to start a mock HTTP server from recorded fixtures so that my integration tests can run deterministically against captured API responses without network access.

## Acceptance Criteria

- [ ] `Server` type in `server.go` implements `http.Handler`
- [ ] Receives HTTP requests and matches them against stored `Tape` fixtures using a `Matcher`
- [ ] Returns the matched fixture's response: status code, headers, and body
- [ ] Configurable fallback status code when no fixture matches (default: 502 Bad Gateway)
- [ ] Constructor accepts a `Store` and `Matcher` via functional options
- [ ] Unit tests in `server_test.go` using `net/http/httptest`
- [ ] Tests cover: exact match response, no-match fallback, custom fallback status, multiple fixtures
- [ ] All exported types and functions have godoc comments
- [ ] No external dependencies — stdlib only
- [ ] `go test -race` passes

## Out of Scope

- Advanced matching (regex, fuzzy) — covered in Milestone 4
- Response templating — Future
- Stateful scenarios — Future
- Sanitization — covered in Milestone 2

Issue #1, opened 2026-03-30
1.1 Core types and storage interface


tibtof:
## Problem Statement

httptape needs foundational types to represent recorded HTTP interactions and a storage abstraction to persist and retrieve them. Without these core types, no other functionality (recording, replay, sanitization) can be built.

## User Story

As a Go developer, I want well-defined types for HTTP request/response pairs and a pluggable storage interface so that I can record, store, and retrieve HTTP interactions using either in-memory or filesystem-based storage.

## Acceptance Criteria

- [ ] `Tape` type defined in `tape.go` representing a recorded HTTP interaction (request + response pair)
- [ ] `RecordedReq` type defined in `tape.go` with method, URL, headers, body, and body hash
- [ ] `RecordedResp` type defined in `tape.go` with status code, headers, and body
- [ ] `Store` interface defined in `store.go` with `Save(Tape) error`, `Load(id string) (*Tape, error)`, `List(filter) ([]Tape, error)`, `Delete(id string) error`
- [ ] `MemoryStore` implementation in `store_memory.go` (safe for concurrent use)
- [ ] `FileStore` implementation in `store_file.go` storing JSON files on disk
- [ ] FileStore directory structure: `fixtures/<route>/<method>_<path_hash>.json`
- [ ] All core types are pure data structures with zero I/O dependencies
- [ ] Unit tests for `MemoryStore` in `store_memory_test.go` covering Save, Load, List, Delete
- [ ] Unit tests for `FileStore` in `store_file_test.go` covering Save, Load, List, Delete
- [ ] All exported types and functions have godoc comments
- [ ] No external dependencies — stdlib only
- [ ] `go vet` and `go test -race` pass

## Out of Scope

- Recording logic (covered in 1.2)
- Replay/mock server (covered in 1.3)
- Request matching (covered in 1.4)
- Sanitization (covered in Milestone 2)
- Functional options for Store constructors (covered in 1.2 alongside Recorder options)

PR #272, opened 2026-05-26
chore(release): dispatch httptape-js after tag publish

tibtof:
## Summary

Append a final step to \`release.yml\` that fires a \`repository_dispatch\` event on \`httptape/httptape-js\` after the GitHub Release is created. The receiver workflow there (shipped in [httptape-js#3](https://github.com/httptape/httptape-js/pull/3)) builds the Go binary for 5 platforms, sets the matching version on every workspace package, and publishes the binary packages + \`vite-plugin-httptape\` to npm.

Closes the dispatcher half of the cross-repo release pipeline locked in [httptape-js ADR-1](https://github.com/httptape/httptape-js/issues/1#issuecomment-4537030867).

## Manual prerequisites (you, before the next release)

| | |
|---|---|
| 1. **Claim \`@httptape\` npm scope** | \`npm org create httptape\` (or via npmjs.com UI) so \`@httptape/binary-*\` can be published. |
| 2. **Reserve \`vite-plugin-httptape\` name** | One-time \`npm publish\` of a placeholder, OR rely on the workflow's first run to claim it. |
| 3. **Mint \`NPM_TOKEN\`** | Automation token with publish rights on \`@httptape/*\` and \`vite-plugin-httptape\`. Add as secret on \`httptape/httptape-js\`. |
| 4. **Mint \`HTTPTAPE_JS_DISPATCH_TOKEN\`** | Fine-grained PAT with \`Actions: write\` on \`httptape/httptape-js\`. Add as secret **on this repo**. |

Until secret #4 is added, the new dispatch step fails loudly but the GitHub Release itself still ships normally — so this PR is safe to merge before the secret is wired.

## Test plan

- [x] YAML parses (\`python3 -c "import yaml; yaml.safe_load(...)"\`)
- [x] Step uses \`GITHUB_REF_NAME#v\` to strip the leading \`v\` so payload version matches npm version
- [x] Receiver workflow in httptape-js has the matching \`event-type: httptape-release\` and reads \`client_payload.version\`
- [ ] First real release after secrets are added fires the dispatch successfully (validated end-to-end by cutting v0.13.2 or a v0.13.2-rc.0)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,18 @@
 # .github/workflows/release.yml
 #
-# Run tests and publish a GitHub Release with auto-generated notes.
-# httptape is a Go library — no binaries needed.
+# Run tests, publish a GitHub Release with auto-generated notes, and notify
+# httptape/httptape-js (the npm-distribution sibling repo) so it can build
+# and publish per-platform binary packages plus vite-plugin-httptape.
+#
 # Triggers: semver tags (v*).
+#
+# Required secrets:
+#   HTTPTAPE_JS_DISPATCH_TOKEN — fine-grained PAT (or GitHub App token) with
+#     "Actions: write" on httptape/httptape-js. Used by the final step to
+#     fire a repository_dispatch event.
+#     If missing, the dispatch step fails loudly and the rest of the release
+#     still ships; npm packages will not be published automatically until
+#     the secret is added.
 
 name: Release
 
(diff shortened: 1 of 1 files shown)


tibtof (commented):
**Reviewer Agent: APPROVE** (cannot submit as formal approval — same GitHub user)

Reviewed the release.yml diff against all seven verification points:

1. **Existing steps untouched** — Checkout, Set up Go, Run tests, Create GitHub Release are byte-identical to main.
2. **Tag-to-version stripping** — `${GITHUB_REF_NAME#v}` correctly strips the leading `v` (e.g. `v0.13.2` -> `0.13.2`), matching what the httptape-js receiver expects in `client_payload.version`.
3. **Dispatch step** — `peter-evans/repository-dispatch@v3`, correct secret name (`HTTPTAPE_JS_DISPATCH_TOKEN`), correct target repo (`httptape/httptape-js`), correct event-type (`httptape-release`), payload carries both `version` and `sha`.
4. **Top-of-file comment** — clearly documents the new behavior, required secret, and failure mode.
5. **Failure semantics** — dispatch is the last step; if it fails, GitHub Release is already created. No half-released state.
6. **No `secrets:` block** — inline `${{ secrets.NAME }}` is the standard pattern.
7. **Permissions** — `contents: write` unchanged, no speculative additions.

Pure CI/YAML change, no library code affected. The `client-payload` uses YAML literal block scalar (`|`) which is the documented pattern for the `repository-dispatch` action.

No issues found. Ready for human review.

PR #294, opened 2026-06-05
fix(caching): include query string and headers in single-flight key

### LINKED ISSUE #285: fix(caching): single-flight key omits query/headers; concurrent waiters get wrong response
## Summary
`CachingTransport`'s single-flight key is `req.Method + "|" + req.URL.Path + "|" + bodyHash` — it **omits the query string and every other dimension the configured `Matcher` may distinguish** (`QueryParamsCriterion`, `HeadersCriterion`, `ContentNegotiationCriterion`). When two concurrent cache-miss requests share method+path+body but differ in query/headers, the second becomes a single-flight *waiter* and receives the *leader's* response — the wrong tape. Only the leader's variant gets recorded.

## Reproduction
With matcher `Method+Path+QueryParams`, two concurrent GETs to `/data?page=1` and `/data?page=2` produced a single upstream hit and both callers got `page=1`'s body.

## Evidence
- `caching_transport.go:339` (key construction) — verified
- waiter path `caching_transport.go:352-383` → `cloneResponse(call.resp, ...)`

## Suggested fix
Include `req.URL.RawQuery` in the key at minimum; ideally derive the key from the same fields the configured matcher uses, or disable single-flight collapsing across distinguishable requests. Document that single-flight assumes the key fully identifies a matchable request.

_Source: audit:2026-06 (bug pass). Reproduced + verified against current code._

tibtof:
Closes #285

## Summary

- Replaces the inline single-flight key `method|path|bodyHash` with a call to the new `singleFlightKey(req, bodyHash)` helper.
- `singleFlightKey` derives the key from: method, URL path, raw query string, body hash, and a canonical SHA-256 hash of all request headers not in `sfHeaderDenylist`.
- `sfHeaderDenylist` excludes hop-by-hop / volatile headers (`Connection`, `Keep-Alive`, `Transfer-Encoding`, `Te`, `Trailer`, `Upgrade`, `Proxy-Connection`, `Content-Length`, `Host`) so transport noise does not needlessly defeat collapsing for genuinely identical requests.
- `roundTripWithSingleFlight` is unchanged — it already accepts `sfKey` as a parameter; only what `RoundTrip` passes changes.
- Updated `WithCacheSingleFlight` godoc to document the key dimensions (AC-7).

## Test plan

- `TestCachingTransport_SingleFlightDistinguishesQueryString` (AC-4): two goroutines hitting `/data?page=1` and `/data?page=2` concurrently; upstream called exactly 2, each goroutine gets its own page body.
- `TestCachingTransport_SingleFlightCollapsesIdenticalRequests` (AC-5): five goroutines hitting the same URL; upstream called exactly 1, all get the same body.
- `TestCachingTransport_SingleFlightDistinguishesHeaders` (AC-2): two goroutines differing only in `Accept`; upstream called exactly 2, each goroutine gets the response for its own `Accept`.
- `TestSingleFlightKey` (table-driven unit test): same key for value-order permutations and denylist-only differences; distinct keys for query string, Accept, and body hash differences.
- All pre-existing `CachingTransport` tests pass unchanged (AC-8).
- `go test -race ./...` passes cleanly.

### DIFF
--- a/caching_transport.go
+++ b/caching_transport.go
@@ -3,9 +3,12 @@ package httptape
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"net/http"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -94,6 +97,15 @@ func WithCacheFilter(fn func(*http.Response) bool) CachingOption {
 // WithCacheSingleFlight controls single-flight deduplication of concurrent
 // cache misses. When true (default), concurrent requests with the same match
 // key share a single upstream call.
+//
+// The single-flight key is derived from: HTTP method, URL path, raw query
+// string, request body hash, and a canonical SHA-256 hash of all request
+// headers except hop-by-hop / volatile headers (see sfHeaderDenylist). This
+// key fully identifies every dimension a Matcher could distinguish, so a
+// single-flight waiter never receives a response that would not match its own
+// request. The cost is reduced collapsing when two equivalent requests differ
+// only in incidental headers (e.g. a per-request trace ID); both then make
+// separate upstream calls.
 //
 // Known limitation: waiters do not observe their request's context cancellation
 // until the leader's upstream call completes. This is a consequence of using
@@ -336,13 +348,80 @@ func (ct *CachingTransport) RoundTrip(req *http.Request) (*http.Response, error)
 	}
 
 	// 6. MISS: forward to upstream.
-	sfKey := req.Method + "|" + req.URL.Path + "|" + bodyHash
+	sfKey := singleFlightKey(req, bodyHash)
 
 	if ct.singleFlight {
 		return ct.roundTripWithSingleFlight(req, reqBody, bodyHash, sfKey)
 	}
 
 	return ct.roundTripUpstream(req, reqBody, bodyHash)
+}
+
+// sfHeaderDenylist names request headers excluded from the single-flight key.
+// These are hop-by-hop / per-connection / volatile headers not inspected by
+// any built-in Matcher; including them would needlessly defeat single-flight
+// collapsing for genuinely-equivalent requests. Keys are stored in canonical
+// (http.CanonicalHeaderKey) form.
+var sfHeaderDenylist = map[string]struct{}{
+	"Connection":        {},
+	"Proxy-Connection":  {},
+	"Keep-Alive":        {},
+	"Transfer-Encoding": {},
+	"Te":                {},
+	"Trailer":           {},
+	"Upgrade":           {},
+	"Content-Length":    {},
+	"Host":              {},
+}
+
+// singleFlightKey derives a single-flight deduplication key that fully
+// identifies a matchable request. It folds in method, URL path, raw query
+// string, the request body hash (already computed by the caller), and a
+// canonical SHA-256 hash of all request headers not in sfHeaderDenylist.
+//
+// The key is safe for any configured Matcher: two requests sharing a key are
+// indistinguishable on every dimension a Matcher could read, so a
+// single-flight waiter never receives a response that would not match its own
+// request.
+//
+// singleFlightKey never fails and performs no I/O. It reads req.Header
+// read-only and never mutates the request.
+func singleFlightKey(req *http.Request, bodyHash string) string {
+	// Collect non-denylisted header names in sorted order.
+	names := make([]string, 0, len(req.Header))
+	for name := range req.Header {
+		canonical := http.CanonicalHeaderKey(name)
+		if _, denied := sfHeaderDenylist[canonical]; !denied {
+			names = append(names, canonical)
+		}
+	}
+	sort.Strings(names)
+
+	// Hash the canonical header representation.
+	h := sha256.New()
+	for _, name := range names {
+		vals := req.Header[name]
+		// Sort a copy of the value slice so multi-valued headers hash
+		// deterministically regardless of insertion order.
+		valsCopy := make([]string, len(vals))
+		copy(valsCopy, vals)
+		sort.Strings(valsCopy)
+
+		// Frame: name \x00 v1 \x01 v2 ... \n — control bytes that cannot
+		// appear in header names or values, keeping serialization unambiguous.
+		h.Write([]byte(name))
+		h.Write([]byte{0x00})
+		for i, v := range valsCopy {
+			if i > 0 {
+				h.Write([]byte{0x01})
+			}
+			h.Write([]byte(v))
+		}
+		h.Write([]byte{0x0a}) // \n
+	}
+	headerDigest := hex.EncodeToString(h.Sum(nil))
+
+	return req.Method + "|" + req.URL.Path + "|" + req.URL.RawQuery + "|" + bodyHash + "|" + headerDigest
 }
 
 // roundTripWithSingleFlight coordinates single-flight dedup for non-SSE
--- a/caching_transport_test.go
+++ b/caching_transport_test.go
@@ -1794,14 +1794,13 @@ func TestCachingTransport_SingleFlightDistinguishesQueryString(t *testing.T) {
 	store := NewMemoryStore()
 
 	var callCount atomic.Int64
-	// ready is closed once both goroutines have entered the upstream stub,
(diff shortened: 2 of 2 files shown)


caching_transport_test.go
```diff
@@ -1780,3 +1780,373 @@ func TestCachingTransport_WithCacheReplayTiming_PreFeatureNoDelay(t *testing.T)
...
+	})
+
+	ct := NewCachingTransport(upstream, store,
+		WithCacheMatcher(NewCompositeMatcher(MethodCriterion{}, PathCriterion{}, QueryParamsCriterion{})),
+		WithCacheSingleFlight(true),
+	)
+
+	// Release the first goroutine once the second has entered the stub.
+	go func() {
+		for arrivedCount.Load() < 2 {
```
tibtof: **Blocker — test hangs (does not fail cleanly) on regression.** I empirically verified this: reverting `sfKey` to the old buggy `method|path|bodyHash` construction makes this test **deadlock for the entire `-timeout` window** (20s locally, the 10-minute default in CI) and then die with a goroutine panic dump — instead of failing with the clear `t.Errorf("upstream called %d times, want 2 ...")` message you wrote.

Why: if keys collapse to one (the regression), only one goroutine reaches the upstream stub. `arrivedCount` never reaches 2, so this spinner never closes `unblock`; the leader stays parked on `<-unblock` and the waiter stays parked on `call.wg.Wait()`. Classic three-way deadlock. The `callCount != 2` assertion is never reached.

Two problems in one:
1. **No timeout guard / deadlock on regression.** Gate the unblock on a bounded wait so a regression fails as an assertion, not a hang. Suggested fix: make the stub signal arrivals over a buffered channel and have the releaser `select` on a `time.After(2*time.Second)` fallback that closes `unblock` and lets the assertions run (they will then fail cleanly with the count message). A regression test that hangs is strictly worse than one that fails.
2. **Busy-spin.** `for arrivedCount.Load() < 2 {}` pegs a CPU core with no `runtime.Gosched()`; even on the happy path it is wasteful under `-race`. Prefer channel-based coordination (a buffered `arrived chan struct{}` the stub sends to, drained by the releaser) over the spin.

Same pattern needs the same fix in `TestCachingTransport_SingleFlightDistinguishesHeaders` (line ~1988).

caching_transport_test.go
```diff
@@ -1780,3 +1780,373 @@ func TestCachingTransport_WithCacheReplayTiming_PreFeatureNoDelay(t *testing.T)
...
+		}, nil
+	})
+
+	ct := NewCachingTransport(upstream, store,
+		WithCacheMatcher(NewCompositeMatcher(MethodCriterion{}, PathCriterion{})),
+		WithCacheSingleFlight(true),
+	)
+
+	go func() {
+		for arrivedCount.Load() < 2 {
```
tibtof: **Blocker (same root cause as the query test).** This header-distinguishing test uses the identical busy-spin + unbounded-block pattern (`for arrivedCount.Load() < 2 {}` → `close(unblock)`; stub parks the first arrival on `<-unblock`). On a key regression it deadlocks for the full test timeout instead of failing on the `callCount != 2` assertion.

Apply the same channel-based, bounded-wait fix here: stub sends to a buffered `arrived` channel; releaser closes `unblock` after both arrive **or** after a short `time.After` fallback so the assertions always run and fail cleanly. Also drops the CPU-pegging spin.

caching_transport.go
```diff
@@ -345,6 +357,73 @@ func (ct *CachingTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+// canonical SHA-256 hash of all request headers not in sfHeaderDenylist.
+//
+// The key is safe for any configured Matcher: two requests sharing a key are
+// indistinguishable on every dimension a Matcher could read, so a
+// single-flight waiter never receives a response that would not match its own
+// request.
+//
+// singleFlightKey never fails and performs no I/O. It reads req.Header
+// read-only and never mutates the request.
+func singleFlightKey(req *http.Request, bodyHash string) string {
```
tibtof, line 389: **Non-blocking — confirmation, no change needed.** The key construction is correct: header-name canonicalization + sort is order-independent; the value-slice is copied before `sort.Strings` so `req.Header` is never mutated; and the `\x00`/`\x01`/`\x0a` framing is unambiguous and non-injectable (none of these bytes are legal in RFC 7230 header names or values). Denylist matches the standard hop-by-hop set + Proxy-Connection, and I confirmed no built-in `Criterion` reads `Host` or `Content-Length`, so excluding them cannot collapse a wrong response for any built-in matcher.

One documented caveat worth keeping in mind (already covered by ADR-47, not asking for a change): the godoc says "safe for any configured Matcher," but a `CriterionFunc` that reads `req.Host` or `Content-Length` would not be distinguished by this key. ADR-47 consciously accepted this (Host lives on `req.Host`, Content-Length is redundant with the body hash) and noted the revisit-if-a-Host-criterion-appears condition. Fine as-is; flagging only so the reviewer trail records that the "any matcher" claim has this one explicit boundary.

tibtof:

(Formal `--request-changes` is blocked because the PR author and reviewer are the same account; recording the verdict here. Inline comments are posted on the diff.)

One blocker, in the **tests only** — the production fix is correct and should be kept as-is.

**Correct (verified):**
- `singleFlightKey` is sound: canonical header-name sort, value-slice copied before sort (no `req.Header` mutation), unambiguous `\x00`/`\x01`/`\x0a` framing. Key = method | path | rawquery | bodyHash | sha256(non-denylisted headers). Matches ADR-47.
- Denylist is the correct hop-by-hop set + Proxy-Connection; confirmed no built-in Criterion reads Host or Content-Length, so excluding them is safe.
- stdlib-only, `gofmt` clean, `go vet` clean, godoc updated (AC-7). All single-flight tests pass under `-race`. AC-1/2/3/5/6/8 satisfied; the `singleFlightKey` unit table is a nice addition.

**Blocker (AC-4 query test + AC-2 header test):** Both regression tests **deadlock and hang for the full test timeout on a key regression** instead of failing on their `callCount != 2` assertion. Verified empirically — reverting `sfKey` to the buggy form parked the test for the entire `-timeout` and killed it with a goroutine dump, never reaching the assertion. The `for arrivedCount.Load() < 2 {}` releaser never fires when a regression collapses both requests into one upstream call (only one arrival), leaving leader on `<-unblock` and waiter on `wg.Wait()` forever. A regression test that hangs masks the failure as a CI timeout rather than a clear FAIL.

**Required fix:** bounded unblock (channel arrival signal + `time.After` fallback that releases and lets the count assertion fail cleanly), and drop the busy-spin for channel coordination. Apply to both tests. Inline comments give a concrete approach.

Once both regression tests fail-clean instead of hang, this is an approve.

tibtof:
Changes-requested blocker addressed in commit cb90d7c.

**What changed** (tests only — production `singleFlightKey` untouched):

Both `TestCachingTransport_SingleFlightDistinguishesQueryString` and `TestCachingTransport_SingleFlightDistinguishesHeaders` replaced the `arrivedCount atomic.Int64` + busy-spin goroutine pattern with:

- A buffered `arrived chan struct{}` (capacity 2) that upstream stubs send to on every entry — no atomic, no spin, no CPU peg.
- A releaser goroutine that drains two arrivals via `select` + `time.After(3 * time.Second)`. When the fix is correct both arrivals come in microseconds. When a key regression collapses both requests into one upstream call, only one arrival fires; the deadline branch closes `unblock` and logs `"only one upstream arrival before deadline — key regression suspected; unblocking so test can fail on callCount assertion"`, then the test completes in ~3s and prints the clear `callCount != 2` failure instead of hanging for the full `-timeout`.

**Regression verification** — I temporarily reverted `singleFlightKey` to the buggy `method|path|bodyHash` form and ran:
```
go test -run "TestCachingTransport_SingleFlightDistinguishesQueryString|TestCachingTransport_SingleFlightDistinguishesHeaders" -timeout 30s -v
```
Both tests FAILED in 3.01s with the exact expected messages (log line + callCount + body assertions). Then restored the fix and confirmed both PASS in 10ms.

**All suites green:**
- `go build ./...` — clean
- `go vet ./...` — clean  
- `go test ./... -timeout 60s` — PASS
- `go test -race ./... -timeout 120s` — PASS
- `go test -run "...QueryString|...Headers" -timeout 30s` — PASS (10ms each)

PR #295, opened 2026-06-05
feat(sanitizer): redact/fake URL query params and userinfo

### LINKED ISSUE #279: security(sanitizer): query-string & userinfo secrets are never sanitized
## Summary
Secrets passed in the **URL query string** (or userinfo) are written to fixtures in cleartext, even with a fully-configured sanitizer pipeline. This directly defeats the "fixtures are safe to commit to version control" guarantee.

## Evidence
There is no built-in `SanitizeFunc` that touches `RecordedReq.URL`. `RedactHeaders`, `RedactBodyPaths`, `FakeFields`, `FakeFieldsWith` only modify headers and JSON bodies. The full URL (including query) is copied verbatim into the tape and marshaled to disk:
- `sanitizer.go` / `faker.go` — verified: no `url`/`query`/`rawquery` handling exists
- `recorder.go:343`, `caching_transport.go:502`, `proxy.go:184` — URL copied verbatim
- `tape.go:150-175` — URL marshaled to disk

So `?api_key=...`, `?access_token=...`, presigned-URL `?sig=...` land in committed fixtures in cleartext.

## Suggested fix
Add `RedactQueryParams(names...)` / `FakeQueryParams(...)` SanitizeFuncs that parse `t.Request.URL`, redact/fake matching keys, and rewrite the URL. Consider redacting userinfo (`https://user:pass@host`). Wire a `redact_query` config action.

_Source: audit:2026-06 (sanitization pass). Verified against current code._

tibtof:
Closes #279

## Summary

- Adds `RedactQueryParams(names ...string) SanitizeFunc` to `sanitizer.go`: replaces matching URL query parameter values with `[REDACTED]`. With no arguments, uses a built-in default list (`api_key`, `access_token`, `token`, `secret`, `password`, `sig`, `signature`, `X-Amz-Signature`, `X-Goog-Signature`).
- Adds `FakeQueryParams(seed string, names ...string) SanitizeFunc`: replaces matching query parameter values with deterministic fakes (`fake_<8-hex-chars-HMAC>`) reusing the existing `computeHMAC` + `fakeString` helpers.
- Adds `DefaultSensitiveQueryParams() []string`: returns a fresh copy of the default sensitive param list (mirrors `DefaultSensitiveHeaders()` copy semantics).
- Both URL sanitizers **unconditionally strip userinfo** (`user:pass@host` → `[REDACTED]@host`, password cleared) regardless of query param match, closing that credential surface.
- URL is re-encoded only when a value actually changed or userinfo was present (option b per ADR-48), keeping non-matching URLs byte-identical.
- Query param name matching is **case-sensitive** per RFC 3986, documented in godoc.
- Adds `ActionRedactQuery = "redact_query"` and `ActionFakeQuery = "fake_query"` to `config.go` with `Rule.Params []string` field, full `Validate()` enforcement, and `BuildPipeline()` wiring.
- No new dependencies: `net/url` is already stdlib.

## Test plan

- `go build ./...` — passes
- `go vet ./...` — passes
- `go test ./...` — passes (main package coverage: **94.7%**, total: **91.1%**)
- `go test -race ./...` — passes clean

Test matrix covers:
- Matching param redacted / non-matching untouched / multi-value per key
- FakeQueryParams determinism (same seed+value → same fake) and different values differ
- Userinfo stripped (with and without matching query param)
- URL byte-stability when nothing changes
- Silent skip on malformed URL
- Only `Request.URL` is modified; headers/body/response are byte-identical
- Config round-trip (`redact_query` and `fake_query`) including all `Validate()` error cases
- `DefaultSensitiveQueryParams` returns a copy; mutating it does not affect internal state

### DIFF
--- a/sanitizer.go
+++ b/sanitizer.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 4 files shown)


sanitizer.go
```diff
@@ -45,6 +46,197 @@ func DefaultSensitiveHeaders() []string {
...
+
+	// Unconditionally strip userinfo (AC-7: userinfo is always a credential).
+	if u.User != nil {
+		u.User = url.User(Redacted)
+		changed = true
+	}
+
+	// Modify matching query parameter values.
+	if len(names) > 0 {
+		values := u.Query()
```
tibtof: **SECURITY BLOCKER — fail-open query leak.** `u.Query()` calls `url.ParseQuery` internally and **silently discards its error**, dropping any param whose value contains malformed percent-encoding (a literal `%` not followed by two hex digits). Concretely, for `https://host/p?api_key=ab%ZZcd` with `names={"api_key"}`:

- `u.Query()` returns `map[]` (the `api_key` entry is dropped),
- the loop never matches `api_key`, so `changed` stays `false`,
- `sanitizeURL` returns `rawURL` **byte-identical** — the cleartext secret `ab%ZZcd` is written to disk.

This is reachable from the live recorder path, not just hand-edited fixtures: a client sending `?api_key=ab%ZZcd` produces `r.URL.String() == "/p?api_key=ab%ZZcd"` (verified — `URL.String()` emits `RawQuery` verbatim, no re-encoding), which the recorder stores at `recorder.go:343`. On a sanitize-on-write security boundary this must **fail closed**.

Fix: call `url.ParseQuery(u.RawQuery)` explicitly and handle the error — on parse failure, do not pass the URL through unchanged. Either redact the entire raw query string, or treat the whole URL as untrusted and redact it. Do not rely on `u.Query()`'s error-swallowing behavior anywhere on this boundary.

sanitizer.go
```diff
@@ -45,6 +46,197 @@ func DefaultSensitiveHeaders() []string {
...
+				}
+			}
+			values[name] = vals
+		}
+		if changed {
+			u.RawQuery = values.Encode()
+		}
+	}
+
+	if !changed {
```
tibtof, line 189: **SECURITY — fragment leak vector (flag, weigh against spec).** `sanitizeURL` never inspects the URL fragment. A URL like `https://host/p?api_key=SECRET#api_key=SECRET` has the query redacted but the fragment retains `SECRET` verbatim on disk.

AC-2 explicitly says "fragment [is] preserved unchanged," so this is per-spec — but on a sanitize-on-write *security boundary* it is still a real leak surface (OAuth implicit flow puts `access_token`/`id_token` in the fragment; some presigned schemes do too). Fragments are also not sent to the server on the wire, so in the pure recorder path they will usually be empty — but the `SanitizeFunc` is public API applied to arbitrary tapes.

Not necessarily a blocker on its own given the spec wording, but please either: (a) extend redaction to scan the fragment for the same param names, or (b) document the fragment-is-not-sanitized caveat prominently in the godoc and the ADR so callers know fragments are out of scope. Right now neither the godoc nor the ADR mentions it.

sanitizer.go
```diff
@@ -45,6 +46,197 @@ func DefaultSensitiveHeaders() []string {
...
+	// Modify matching query parameter values.
+	if len(names) > 0 {
+		values := u.Query()
+		for name, vals := range values {
+			if _, ok := names[name]; !ok {
+				continue
+			}
+			for i, v := range vals {
+				vals[i] = replace(v)
+				if vals[i] != v {
```
tibtof: **Minor — `changed` flag can miss a same-value match (cosmetic, not a leak).** The `changed` flag is only set when `vals[i] != v`. If a sensitive param's value already equals its replacement (e.g. `?api_key=[REDACTED]` for redact, or an input that hashes to the identical fake), `changed` stays false and the URL is returned byte-identical. This is harmless for redact (already redacted) and astronomically unlikely for fake.

However, combined with option (b), it means "this param was a sensitive match" is conflated with "a byte changed." Consider tracking `matched` separately from `changed`, or unconditionally setting `changed = true` whenever a configured name is found in the query (regardless of whether the value differs). That makes the security intent — "a sensitive param was present, so re-encode authoritatively" — explicit and removes any dependence on value comparison. Non-blocking, but it tightens the boundary.

tibtof:
## Review: CHANGES REQUESTED

Strong PR overall — the structure mirrors the existing header/body sanitizers exactly, godoc is thorough, the test matrix is broad, userinfo handling is correct on both match and no-match paths, determinism reuses the canonical `computeHMAC`+`fakeString` path, `DefaultSensitiveQueryParams` returns a defensive copy, and the config Validate/BuildPipeline/DisallowUnknownFields wiring is all correct. Byte-stability (option b) is implemented as specified and does not skip userinfo stripping.

But this is the sanitize-on-write security boundary, and I found one **fail-open leak that lets a cleartext secret reach disk** — blocking.

### Blocker

1. **`u.Query()` swallows `ParseQuery` errors → cleartext secret persisted (`sanitizer.go:108`).** When a sensitive param's value contains malformed percent-encoding (a literal `%` not followed by two hex digits, e.g. `?api_key=ab%ZZcd`), `u.Query()` silently drops that param. The redaction loop never matches it, `changed` stays false, and `sanitizeURL` returns the raw URL byte-identical — secret intact. Verified reachable from the live recorder path: a client sending `?api_key=ab%ZZcd` yields `r.URL.String() == "/p?api_key=ab%ZZcd"` (URL.String emits RawQuery verbatim), stored at `recorder.go:343`. On this boundary the query must **fail closed** on a parse error (redact the whole query / whole URL), never pass through. Use `url.ParseQuery(u.RawQuery)` with explicit error handling instead of `u.Query()`. Please add a regression test (malformed-encoding sensitive value → secret absent from output).

### Should address

2. **Fragment is never sanitized (`sanitizer.go:126`).** Per-spec (AC-2 says fragment is preserved), but it is a real leak surface (OAuth implicit-flow tokens live in the fragment) and neither the godoc nor ADR-48 documents that fragments are out of scope. Either scan the fragment for the same names, or document the caveat prominently.

### Nice to have (non-blocking)

3. **Decouple `matched` from `changed` (`sanitizer.go:115`).** Set `changed = true` whenever a configured param name is found, not only when a byte differs, so the security intent ("a sensitive param was present, re-encode authoritatively") doesn't depend on value comparison.

Once #1 is fixed with a regression test, and #2 is addressed (fix or documented), I'm happy to approve.

tibtof:
All three issues addressed in commit f6c0b00. Summary:

**1. BLOCKER fixed — fail-closed on malformed query string**

Replaced `u.Query()` with an explicit `url.ParseQuery(u.RawQuery)` call with error handling. On parse failure (e.g. `?api_key=ab%ZZcd`), the entire `RawQuery` is replaced with `[REDACTED]` — the cleartext query string is never returned or persisted. Userinfo stripping still happens before the query parse attempt, so it is not leaked back. Regression tests added:
- `TestRedactQueryParams_MalformedPercentEncodingSecretDoesNotReachPersistedURL` (table-driven, 3 malformed-encoding variants)
- `TestFakeQueryParams_MalformedPercentEncodingSecretDoesNotReachPersistedURL`
- `TestRedactQueryParams_MalformedQueryStringStillStripsUserinfo`

**2. SHOULD ADDRESS done — fragment params are now sanitized**

Fragment values whose param name appears in the sensitive set are redacted/faked. On fragment parse error the entire fragment is replaced with `[REDACTED]`. Documented in godoc for both `RedactQueryParams` and `FakeQueryParams`. Note: Go's `url.Parse` itself rejects malformed percent-encoding in fragments (unlike in query strings where it tolerates it), so the fragment-parse-error path in `sanitizeURL` is a defensive guard rather than a reachable production path — documented in the test. Tests added:
- `TestRedactQueryParams_SensitiveParamInFragmentIsRedacted`
- `TestRedactQueryParams_NonSensitiveParamInFragmentIsUntouched`
- `TestRedactQueryParams_WellFormedFragmentWithKeyValueIsRedacted`
- `TestRedactQueryParams_MalformedFragmentURLFailsStructuralParse` (documents the Go url.Parse behavior)
- `TestFakeQueryParams_SensitiveParamInFragmentIsFaked`

**3. NON-BLOCKING done — changed decoupled from value equality**

`changed = true` is now set whenever a configured sensitive param name is found in the query, not only when the replacement value differs byte-for-byte. Tests added:
- `TestRedactQueryParams_SensitiveParamAlreadyRedactedCausesReencode`

All previously approved behavior preserved: determinism, userinfo stripping, defensive copy, byte-stability for untouched URLs, config Validate/BuildPipeline. `go test ./...`, `go test -race ./...`, `go build ./...`, `go vet ./...` all pass. Coverage 94.7%.

PR #297, opened 2026-06-05
feat(cli): record/proxy fail closed with safe default sanitization

### LINKED ISSUE #280: security(cli): record/proxy fail open without --config (no raw recording mode is violated)
## Summary
The CLI `record` and `proxy` subcommands apply **no sanitizer at all** unless `--config` is passed — they fall back to the default no-op pipeline and record Authorization headers, cookies, bodies, and URLs raw to disk. CLAUDE.md states "there is no raw recording mode," but the CLI default is effectively raw. The CLI/Docker image is the documented channel for non-Go users, who are the most likely to run without a config and assume sanitize-on-write is on.

## Evidence
- `cmd/httptape/main.go:453-462` (record) and `:603-610` (proxy): sanitizer only added when `*configPath != ""` — verified.
- Defaults fall back to no-op `NewPipeline()` (`recorder.go:220`, `proxy.go:473`).

## Suggested fix
Fail closed: either require `--config`, or auto-apply `RedactHeaders()` (default sensitive headers) when no config is supplied, **and** print a prominent startup warning. At minimum, warn loudly when recording with a no-op pipeline.

_Source: audit:2026-06 (sanitization pass). Verified against current code._

tibtof:
Closes #280

## Summary

- Adds `--unsafe-raw` bool flag to both `record` and `proxy` subcommands.
- When neither `--config` nor `--unsafe-raw` is supplied (the default), auto-applies `NewPipeline(RedactHeaders(), RedactQueryParams())` and prints `safeDefaultWarning` to stderr — so sensitive headers (`Authorization`, `Cookie`, `Set-Cookie`, `X-Api-Key`, `Proxy-Authorization`, `X-Forwarded-For`) and query params (`api_key`, `access_token`, `token`, `secret`, `password`, `sig`, `signature`, `X-Amz-Signature`, `X-Goog-Signature`) are redacted before touching disk.
- `--config <path>` continues to replace the default pipeline exactly as before; no default pipeline is prepended/appended, and the safe-default warning is suppressed.
- `--unsafe-raw` applies the library no-op pipeline and prints `unsafeRawWarning` (contains `UNSAFE`) to stderr.
- `--config` + `--unsafe-raw` together is a usage error (exit 1), caught after `fs.Parse` but before any file read.
- For `proxy`, the safe default is passed via `WithProxySanitizer` (same option `--config` uses), so it lands on the L2 persisted write path.
- No library changes; CLI-only fix per ADR-49.

## Test plan

13 new tests in `cmd/httptape/main_test.go`, all race-clean (`go test -race ./...`):

- `TestRecordFailClosed_DefaultRedactsAuthAndToken` — fixture has `Authorization: [REDACTED]` and `?token=[REDACTED]`
- `TestRecordFailClosed_DefaultWarningOnStderr` — stderr contains `WARNING`, `Authorization`, `api_key`
- `TestRecordFailClosed_ConfigSuppressesDefaultWarning` — `safeDefaultWarning` absent when `--config` supplied
- `TestRecordFailClosed_UnsafeRawRecordsRaw` — fixture has literal `Bearer secret`; stderr contains `UNSAFE`
- `TestRecordFailClosed_UnsafeRawPlusConfigExitsUsage` — exit code 1
- `TestRecordHelpListsUnsafeRaw` — `unsafe-raw` in `record -h` output
- Symmetric `TestProxyFailClosed_*` tests for all of the above
- `TestProxyHelpListsUnsafeRaw`
- `TestSafeDefaultWarningDriftGuard` — asserts every element of `DefaultSensitiveHeaders()` and `DefaultSensitiveQueryParams()` appears in `safeDefaultWarning`

Verification:
```
go build ./...
go vet ./...
go test ./...
go test -race ./...
```
All pass.

### DIFF
--- a/cmd/httptape/main_test.go
+++ b/cmd/httptape/main_test.go
@@ -15,6 +15,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/httptape/main_test.go
```diff
@@ -1358,3 +1359,517 @@ func generateTestCertAndKey(t *testing.T, hosts ...string) (certPEM, keyPEM []by
...
+		if err != nil {
+			panic(err)
+		}
+		req.Header.Set("Authorization", "Bearer secret")
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return // server may be shutting down; tolerate connection error
+		}
+		resp.Body.Close()
+		_ = upstream
```
tibtof, line 1529: **Minor (non-blocking)** — dead parameter. `makeProbeWithAuthAndToken` takes an `*httptest.Server` it never uses; the `_ = upstream` on this line exists only to silence the unused-param. The probe only needs `baseURL`. Either drop the `upstream` parameter entirely (callers already `defer upstream.Close()` in the test bodies), or use it. Not a blocker, but it reads as leftover scaffolding.

Issue #315, opened 2026-06-11
fix(caching): SSE single-flight waiters re-query before stream completes — false store-write errors + N duplicate tapes


tibtof:
## Summary
For SSE, single-flight is effectively a no-op that also produces noise and duplicate fixtures. The leader marks the call done at **header time** (`call.wg.Done()` in the SSE branch, `caching_transport.go` `roundTripWithSingleFlight`), but the SSE tape is persisted only at **stream completion**. Waiters wake immediately, `reQueryStoreForSSE` does a **single** store query (no wait/poll), finds nothing for any stream that doesn't complete near-instantly, then:

1. fires `onErrorSafe("...no matching tape found after leader completed...")` — a misleading false alarm ("leader's store write may have failed" — it hasn't happened yet) on **every** concurrent SSE waiter;
2. falls through to a direct upstream call **with its own recording tee** — so N concurrent identical SSE misses produce **N upstream calls and N duplicate tapes** persisted at completion.

The in-code comment claims "waiters that arrive later will find the tape in the store (once the stream completes)" — that only holds for near-instant streams.

## Consequences
- Duplicate SSE fixtures pollute the store; later matches between equal-scoring duplicates are then arbitrary (see the nondeterministic tie-breaking issue).
- onError noise makes real store failures hard to spot.

## Suggested fix
Be honest about semantics: for SSE misses, skip single-flight registration entirely (waiters go upstream without the false-alarm path), drop the misleading error, and document that single-flight collapses only buffered responses. Optionally dedupe at save time (skip persisting when a matching SSE tape landed since the request started).

_Source: audit 2026-06-11, hands-on cross-component pass (Fable 5). Code-path traced through roundTripWithSingleFlight → reQueryStoreForSSE._

PR #335, opened 2026-08-05
fix(#300): upstream timeout no longer cancels live SSE streams

### LINKED ISSUE #300: fix(caching): WithCacheUpstreamTimeout cancels live SSE streams on return
## Summary
`WithCacheUpstreamTimeout` wraps the request context with `context.WithTimeout` and `defer cancel()` in `roundTripUpstream` (`caching_transport.go:530-534`). For buffered (non-SSE) responses the body is fully read before return, so this is safe. But for SSE, `roundTripSSE` returns the **live streaming body** — and the deferred `cancel()` fires the moment `roundTripUpstream` returns, making `http.Transport` tear down the response body while the caller is still reading it.

## Reproduction (from audit)
SSE upstream + `WithCacheUpstreamTimeout(30*time.Second)` → caller reads 0 bytes, gets `context canceled`, and the tape is discarded as truncated. Any embedder combining single-flight + upstream-timeout with SSE recording is fully broken.

## Evidence
- `caching_transport.go:530-534` — `ctx, cancel := context.WithTimeout(...); defer cancel()`
- `caching_transport.go:556-557` — SSE path returns the streaming body from inside the same function

## Suggested fix
On the SSE path, don't cancel on return: tie cancellation to stream completion (hand `cancel` to the SSE reader's `Close`/`onDone`), or skip the timeout wrapper for SSE and bound only the header phase.

_Source: audit:2026-06 (2026-06-11 bug pass). Reproduced; code path re-verified before filing._

tibtof:
Closes #300

## Root cause

`roundTripUpstream` in `caching_transport.go` wrapped the request with
`context.WithTimeout` and `defer cancel()`. Two bugs for the SSE path:

1. `defer cancel()` fires when `roundTripUpstream` returns. For SSE,
   `roundTripSSE` returns a *live* streaming body, so the deferred cancel
   runs immediately, tearing the body down while the caller still reads it.
2. Even without the defer, a `WithTimeout` deadline attached to the transport
   context watches the entire body lifetime. A 60s SSE stream is aborted at 30s
   with a 30s timeout — **and a `WithTimeout` deadline cannot be disarmed after
   creation**, so just moving `cancel` to `Close()` is not sufficient on its own.

## Fix

Replace `context.WithTimeout` + `defer cancel()` with a disarmable timer:

```
ctx, cancel = context.WithCancelCause(req.Context())
timer = time.AfterFunc(ct.upstreamTimeout, func() { cancel(context.DeadlineExceeded) })
```

A top-level deferred cleanup stops the timer and calls `cancel(nil)` on return.

**Buffered path**: unchanged semantics. The timer stays armed through
`io.ReadAll`, so a stalled body read is still aborted as
`context.DeadlineExceeded` (net/http uses `context.Cause` in Go 1.21+, so the
cause passed to `cancel` surfaces to the caller correctly).

**SSE path**: when SSE headers arrive, the timer is stopped (`timer.Stop()`)
and cancel ownership transfers to the new `cancelOnCloseReadCloser` type
wrapping the response body. `cancelOnCloseReadCloser.Close()` calls
`cancel(nil)` exactly once (via `sync.Once`), releasing the upstream context
when the caller is done reading — not when `roundTripUpstream` returns.

## Files changed

- `caching_transport.go` — `roundTripUpstream`, `roundTripSSE` (new `cancel`
  parameter), `cancelOnCloseReadCloser`, `WithCacheUpstreamTimeout` godoc.
- `caching_transport_test.go` — four new behavior-named test cases.

## Test plan

New tests in `TestCachingTransport_UpstreamTimeoutSSEvsBuffered`:

- **"SSE stream stays open past upstream timeout while streaming"** — httptest
  server emits 4 events at 40ms intervals (160ms total) with a 60ms timeout.
  Asserts all events are readable with no error, and the persisted tape has all
  4 events. This is the core regression guard.

- **"upstream timeout aborts an SSE request that stalls before sending headers"**
  — server blocks before writing headers. Asserts `errors.Is(err, context.DeadlineExceeded)`.

- **"upstream timeout still bounds a buffered response whose body stalls"** — mock
  transport returns a pipe-backed body that stalls until the context is cancelled.
  Asserts `onError` is called with `context.DeadlineExceeded` (the internal body
  read fails; the response is still returned with empty body as before).

Plus **`TestCachingTransport_ClosingSSEBodyReleasesUpstreamContext`** — captures
the request context, drives a short SSE stream to completion, asserts context is
still live before `Close()` and cancelled after. Proves cancel ownership
transferred and the context does not leak.

All tests pass: `go test ./... -race -count=1` and `go test -run Timeout -race -count=5 ./...`.

### DIFF
--- a/caching_transport_test.go
+++ b/caching_transport_test.go
@@ -2175,3 +2175,244 @@ func TestSingleFlightKey(t *testing.T) {
 		})
 	}
 }
+
+// ---------------------------------------------------------------------------
+// Upstream timeout: SSE vs buffered path correctness (#300)
+// ---------------------------------------------------------------------------
+
+// TestCachingTransport_UpstreamTimeoutSSEvsBuffered verifies that the
+// disarmable-timer approach (context.WithCancelCause + time.AfterFunc) behaves
+// correctly for both SSE and buffered responses.
+func TestCachingTransport_UpstreamTimeoutSSEvsBuffered(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		run  func(t *testing.T)
+	}{
+		{
+			// Core regression: the SSE stream must stay readable past the timeout.
+			// Previously, defer cancel() fired on return from roundTripUpstream,
+			// tearing down the body immediately.
+			name: "SSE stream stays open past upstream timeout while streaming",
+			run: func(t *testing.T) {
+				t.Parallel()
+				store := NewMemoryStore()
+
+				const timeout = 60 * time.Millisecond
+				const eventInterval = 40 * time.Millisecond
+				const numEvents = 4 // span = 4 * 40ms = 160ms >> timeout
+
+				srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					w.Header().Set("Content-Type", "text/event-stream")
+					w.WriteHeader(200)
+					flusher, ok := w.(http.Flusher)
+					if !ok {
+						t.Error("httptest server does not support Flusher")
+						return
+					}
+					for i := 0; i < numEvents; i++ {
+						fmt.Fprintf(w, "data: event-%d\n\n", i)
+						flusher.Flush()
+						if i < numEvents-1 {
+							time.Sleep(eventInterval)
+						}
+					}
+				}))
+				defer srv.Close()
+
+				ct := NewCachingTransport(http.DefaultTransport, store,
+					WithCacheUpstreamTimeout(timeout),
+					WithCacheSingleFlight(false),
+				)
+
+				req, _ := http.NewRequest("GET", srv.URL+"/stream", nil)
+				resp, err := ct.RoundTrip(req)
+				if err != nil {
+					t.Fatalf("RoundTrip error: %v", err)
+				}
+
+				body, readErr := io.ReadAll(resp.Body)
+				resp.Body.Close()
+
+				if readErr != nil {
+					t.Fatalf("SSE body read error after timeout window: %v", readErr)
+				}
+
+				for i := 0; i < numEvents; i++ {
+					want := fmt.Sprintf("event-%d", i)
+					if !strings.Contains(string(body), want) {
+						t.Errorf("SSE body missing %q; got %q", want, string(body))
+					}
+				}
+
+				// Wait for the async onDone callback to persist the tape.
+				time.Sleep(150 * time.Millisecond)
+
+				tapes, _ := store.List(context.Background(), Filter{})
+				if len(tapes) != 1 {
+					t.Fatalf("store has %d tapes after SSE stream, want 1", len(tapes))
+				}
+				if !tapes[0].Response.IsSSE() {
+					t.Error("stored tape is not SSE")
+				}
+				if len(tapes[0].Response.SSEEvents) != numEvents {
+					t.Errorf("stored tape has %d SSE events, want %d", len(tapes[0].Response.SSEEvents), numEvents)
+				}
+			},
+		},
+		{
+			// The header phase must still be bounded: an SSE request that never
+			// sends headers should time out.
+			name: "upstream timeout aborts an SSE request that stalls before sending headers",
+			run: func(t *testing.T) {
+				t.Parallel()
+				store := NewMemoryStore()
+
+				srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					// Block until the client disconnects (timeout fires).
+					<-r.Context().Done()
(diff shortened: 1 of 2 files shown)


caching_transport_test.go
```diff
@@ -2175,3 +2175,244 @@ func TestSingleFlightKey(t *testing.T) {
...
+
+				for i := 0; i < numEvents; i++ {
+					want := fmt.Sprintf("event-%d", i)
+					if !strings.Contains(string(body), want) {
+						t.Errorf("SSE body missing %q; got %q", want, string(body))
+					}
+				}
+
+				// Wait for the async onDone callback to persist the tape.
+				time.Sleep(150 * time.Millisecond)
```
tibtof, line 2250: **Non-blocking (test robustness)**: This is a fixed `time.Sleep(150ms)` followed by a hard `t.Fatalf` on `len(tapes) != 1`. The assertion depends on the async `onDone` callback having persisted the tape within that window. It passed 5x under `-race` here, but sleep-then-assert on async completion is a classic CI flake source under load. Prefer polling the store with a deadline (e.g. loop up to ~2s, 10ms steps, break when `len(tapes)==1`) so the test waits exactly as long as needed and no longer. Lower-bound waiting is fine per the project bar; this just makes it non-flaky.

caching_transport_test.go
```diff
@@ -2175,3 +2175,244 @@ func TestSingleFlightKey(t *testing.T) {
...
+				t.Parallel()
+				store := NewMemoryStore()
+
+				var (
+					capturedErr error
+					errMu       sync.Mutex
+				)
+
+				upstream := roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+					pr, pw := io.Pipe()
```
tibtof, line 2313: **Non-blocking (coverage note)**: This buffered-stall case uses a mock roundTripper + `io.Pipe` closed with `context.Cause`, so it proves the buffered branch keeps the timer armed and routes the error to `onError` — good and deterministic. But it does not exercise real `net/http` body-read cancellation, so the *error-identity through net/http on a stalled buffered body read* isn't directly covered here. The header-phase identity IS covered by the real-server sibling ('...stalls before sending headers'). Since the whole point of the mechanism swap (WithTimeout → WithCancelCause+AfterFunc) is that `ctx.Err()` is now `Canceled` while `Cause` is `DeadlineExceeded`, a real-server buffered-body-stall asserting `errors.Is(err, context.DeadlineExceeded)` would close the last gap. Optional.

tibtof (commented):
VERDICT: APPROVE (recorded as a comment — GitHub blocks self-approval since this account authored the PR).

Reviewed the disarmable-timer approach (`context.WithCancelCause` + `time.AfterFunc` replacing `context.WithTimeout`+`defer cancel()`) against the approved design on #300.

**Design conformance**: exact. Timer bounds the header phase for every response; stays armed through the buffered body read; is stopped the instant SSE headers arrive, with cancel ownership handed to `cancelOnCloseReadCloser` which fires `cancel(nil)` once on body Close.

**Correctness**:
- Cancel is guaranteed on every path. Transport error and all buffered exits (cacheFilter reject, read error, save error) keep `transferred=false` → deferred `cancel(nil)` matches the original `defer cancel()` timing (after ReadAll + Save). The SSE branch sets `transferred=true` and `roundTripSSE` is linear with no early return, so the wrapper is always installed when `cancel != nil`.
- No leak from the `transferred` logic: when `cancel == nil` (`upstreamTimeout == 0`) there is nothing to release.
- `transferred` is read/written only in `roundTripUpstream`'s own goroutine (incl. its deferred func) — no race.
- Double-cancel safe: `CancelCauseFunc` is idempotent (first cause wins); the timer-vs-Stop / `cancel(nil)` overlap cannot panic.

**Error identity**: header-phase timeout surfaces `context.DeadlineExceeded` through net/http (Transport uses `context.Cause`), verified by the real-server 'stalls before sending headers' test.

**Buffered semantics unchanged**; **#315 untouched** (only `caching_transport.go` + its test changed; single-flight `wg.Done` timing and the SSE waiter re-query path are unmodified; the leader's wrapped body is transparent since waiters go through `reQueryStoreForSSE`).

**Quality**: gofmt + `go vet` clean; `go test ./... -race -count=1` green; new timing tests passed 5x under -race with no flakes and assert no upper bounds on elapsed time; godoc updated on `WithCacheUpstreamTimeout`, `roundTripSSE`'s new param, and `cancelOnCloseReadCloser`.

Two non-blocking inline notes: (1) prefer polling over the fixed 150ms sleep before the tape assertion; (2) optional real-server buffered-body-stall test to close the last error-identity gap. Neither blocks merge.

PR #336, opened 2026-08-05
fix(#306): matcher-only --config no longer disables sanitization

### LINKED ISSUE #306: security(cli): matcher-only --config silently disables sanitization; warning omits body non-coverage
## Summary
Two fail-open gaps remaining in the #297 (ADR-49) CLI fail-closed design, plus one cosmetic ADR-drift nit. Verified against current code before filing.

### 1. Matcher-only `--config` silently disables ALL sanitization
`Config.Validate` accepts `Rules: []` when a `matcher` block is present (`config.go:149-151`). In `runRecord`/`runProxy`, ANY `--config` takes the first switch arm (`cmd/httptape/main.go:485-498`, `:647-660`): `BuildPipeline()` over zero rules → no-op pipeline, and **neither warning is printed**. A user who passes a matcher-only config (reasonable — it's the same config file `serve` uses for matching) gets raw traffic, including Authorization headers and URL credentials, written to disk with zero warning. This is exactly the failure mode #280 was meant to close.
**Fix:** if `cfg.Rules` is empty in record/proxy, layer in `defaultCLISanitizer()` (or refuse / warn loudly).

### 2. `safeDefaultWarning` doesn't disclose that bodies are NOT redacted
`defaultCLISanitizer()` is `RedactHeaders() + RedactQueryParams()` only (`main.go:74-79`) — no body rules (generic body redaction needs field paths, so this is by design). But the warning (`main.go:58-62`) enumerates only what IS redacted; a user would reasonably believe the "safe default" covers `{"password":"hunter2"}` in a login POST or an `access_token` in a token-exchange response. It does not — those persist verbatim.
**Fix:** append to `safeDefaultWarning`: "Request/response BODIES are NOT redacted by default — use --config with redact_body/fake rules for secrets in bodies." Mirror in docs (tracked separately).

### 3. (Minor) mutual-exclusion check runs after disk writes, contradicting ADR-49
ADR-49 claims the `--config`+`--unsafe-raw` error path "must not touch disk", but the check runs after `NewFileStore` (which `os.MkdirAll`s the fixtures dir) and after `BuildTLSConfig` (reads PEM files when TLS flags set) — `main.go:455-482` (record), `:618-644` (proxy). Move the check earlier or soften the ADR wording.

_Source: audit:2026-06 (2026-06-11 bug + sanitization + architecture passes). Items 1-2 verified directly._

tibtof:
Closes #306

## Summary

- **Fail closed on matcher-only configs**: when `--config` is supplied but contains no sanitization rules (`len(cfg.Rules) == 0`), `record` and `proxy` now apply `defaultCLISanitizer()` and print `emptyRulesConfigWarning` instead of silently falling through to a no-op pipeline. Configs with non-empty rules are still used verbatim (ADR-49 clause 1 preserved).
- **Disclose body non-coverage**: both `safeDefaultWarning` and `emptyRulesConfigWarning` now share a `safeDefaultCoverage` constant that explicitly states request/response BODIES are NOT redacted by the header+query default, closing the false-assurance gap.
- **Mutual-exclusion check before disk writes**: the `--config`/`--unsafe-raw` conflict check is moved before `NewFileStore` (`os.MkdirAll`) and `BuildTLSConfig` (PEM reads) in both `runRecord` and `runProxy`, so the error path never touches disk as ADR-49 literally requires.
- **ADR-49 amendment**: appended to `decisions.md` documenting all three changes. `decisions.md` is committed in this PR.

## Test plan

- `TestRecordFailClosed_MatcherOnlyConfigRedactsAuthAndToken` / `TestProxyFailClosed_MatcherOnlyConfigRedactsAuthAndToken` — live probe with matcher-only config; asserts Authorization and token are `[REDACTED]` on disk.
- `TestRecordFailClosed_MatcherOnlyConfigWarnsSafeDefault` / proxy twin — asserts `emptyRulesConfigWarning` substring appears on stderr.
- `TestSafeDefaultWarningsDiscloseBodiesNotRedacted` — asserts both warning strings contain `"BODIES are NOT redacted"`.
- `TestSafeDefaultWarningDriftGuard` (extended) — now table-driven over both `safeDefaultWarning` and `emptyRulesConfigWarning`; asserts every `DefaultSensitiveHeaders()` and `DefaultSensitiveQueryParams()` token appears in each.
- `TestRecordUnsafeRawPlusConfigDoesNotTouchDisk` / `TestProxyUnsafeRawPlusConfigDoesNotTouchDisk` — uses a nested non-existent fixtures path; asserts exit is `exitUsage` AND `os.Stat` returns `IsNotExist`, proving disk is untouched.
- `go test ./... -race -count=1` — all green locally.

### DIFF
--- a/cmd/httptape/main.go
+++ b/cmd/httptape/main.go
@@ -52,14 +52,28 @@ const (
 
 var logger = log.New(os.Stderr, "httptape: ", 0)
 
+// safeDefaultCoverage enumerates exactly what the safe default sanitizer
+// redacts — and, critically, what it does NOT (bodies). Shared by both
+// safe-default warnings so the disclosure cannot drift between them.
+const safeDefaultCoverage = "Redacted headers: Authorization, Cookie, Set-Cookie, X-Api-Key, Proxy-Authorization, X-Forwarded-For. " +
+	"Redacted query params: api_key, access_token, token, secret, password, sig, signature, X-Amz-Signature, X-Goog-Signature. " +
+	"URL userinfo (user:password@) always stripped. " +
+	"Request/response BODIES are NOT redacted by default -- use --config with redact_body/fake rules for secrets in bodies."
+
 // safeDefaultWarning is printed to stderr when neither --config nor --unsafe-raw
 // is supplied. It names every header and query param category from the library's
 // default sensitive lists so users know exactly what is being redacted.
 const safeDefaultWarning = "WARNING: no --config supplied; applying safe default sanitization. " +
-	"Redacted headers: Authorization, Cookie, Set-Cookie, X-Api-Key, Proxy-Authorization, X-Forwarded-For. " +
-	"Redacted query params: api_key, access_token, token, secret, password, sig, signature, X-Amz-Signature, X-Goog-Signature. " +
-	"URL userinfo (user:password@) always stripped. " +
-	"To customize, use --config. To disable all sanitization (not recommended), use --unsafe-raw."
+	safeDefaultCoverage +
+	" To customize, use --config. To disable all sanitization (not recommended), use --unsafe-raw."
+
+// emptyRulesConfigWarning is printed when a --config is supplied but contains
+// no sanitization rules (matcher-only or empty "rules"). record/proxy ignore the
+// matcher block, so with zero rules the pipeline would be a no-op; instead we fail
+// closed by applying the safe default, matching the no-config posture.
+const emptyRulesConfigWarning = "WARNING: --config supplied but it contains no sanitization rules (matcher-only or empty \"rules\"); applying safe default sanitization. " +
+	safeDefaultCoverage +
+	" To sanitize with custom rules, add a non-empty \"rules\" array. To disable all sanitization (not recommended), drop --config and use --unsafe-raw."
 
 // unsafeRawWarning is printed to stderr when --unsafe-raw is supplied. It is
 // intentionally loud to make clear that the fail-closed default has been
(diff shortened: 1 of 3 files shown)


tibtof:
**Reviewer verdict: APPROVE** (posted as comment — GitHub blocks self-approval on own PR).

Reviewed the fix for #306 against the approved ADR-49 amendment design.

**Design conformance (verified in BOTH subcommands):**
- runRecord and runProxy both branch on `len(cfg.Rules) == 0` → apply `defaultCLISanitizer()` (real working pipeline: RedactHeaders + RedactQueryParams, not nil/no-op) + print `emptyRulesConfigWarning`. Non-empty rules still use `cfg.BuildPipeline()` verbatim.
- `serve` path and `Config.Validate` untouched.
- `--config`/`--unsafe-raw` check moved before both `NewFileStore` (MkdirAll) and `BuildTLSConfig` (PEM reads) in both functions; old post-store check fully removed, no duplicate.

**Fail-open analysis (security boundary):**
- `len(cfg.Rules) == 0` is provably the exact no-op case: every action in `BuildPipeline` appends a non-nil SanitizeFunc, and `RedactHeaders()`/`RedactQueryParams()` with empty args default to the sensitive lists — so even a lone "empty" redact rule still redacts defaults. No remaining fail-open edge for single-rule misconfigs.
- Invalid configs error at `LoadConfigFile`; `rules:[]` without a matcher fails `Validate`, so only genuine matcher-only configs reach the fail-closed branch.

**Warning accuracy:** shared `safeDefaultCoverage` correctly discloses bodies are NOT redacted; lists and userinfo match `defaultCLISanitizer` behavior.

**Tests:** behavior-named, assert real fixture content (`Authorization` + `token` == `[REDACTED]` on disk), disk-not-touched tests verify the nested fixtures dir is never created on the usage-error path, drift-guard extended to both warnings. `go vet ./...` clean, `go test ./... -race -count=1` passes, gofmt clean.

**ADR-49 amendment** present in decisions.md and factually matches the implementation.

Ready for human review.

PR #341, opened 2026-08-05
fix(#304): record no longer drops buffered recordings on interrupt

### LINKED ISSUE #304: fix(cli): record drops buffered async recordings on SIGINT
## Summary
`runRecord` flushes the recorder only inside the signal-handler goroutine, after `httpServer.Shutdown` (`cmd/httptape/main.go:541-555`). But `ListenAndServe` returns `http.ErrServerClosed` as soon as `Shutdown` is *called*, so the main goroutine returns from `run()` and the process exits while `recorder.Close()` (which drains the async tape channel, `recorder.go:477-511`) is still racing in the background. Tapes sitting in the async buffer at Ctrl-C are silently dropped. `runProxy` is protected by its synchronous `defer tapeProxy.Close()`; `runRecord` has no equivalent.

## Suggested fix
Add `defer recorder.Close()` in `runRecord` (Close is idempotent via `closeOnce`), or have the shutdown goroutine signal completion and wait before returning.

_Source: audit:2026-06 (2026-06-11 bug pass). Code-path verified; net/http Shutdown semantics are documented._

tibtof:
Closes #304

## Summary

- **Root cause** — `recorder.Close()` (which drains the async tape channel) was called only inside the signal-handler goroutine, *after* `httpServer.Shutdown` returned. `ListenAndServe` returns as soon as `Shutdown` is called, so `run()` — and in a real binary, `os.Exit` — completed before the goroutine reached `Close()`. All buffered tapes were silently dropped.

- **Fix in `cmd/httptape/main.go`** — added `defer recorder.Close()` in `runRecord` immediately after the recorder is created. The drain now runs synchronously before `runRecord` returns, guaranteeing every queued tape is on disk before the process exits. The now-redundant `recorder.Close()` call in the signal goroutine is removed and replaced with a comment (Close is idempotent via `sync.Once`, so double-close is safe regardless).

- **Fix in `recorder.go`** — the post-close silent-drop path in `persistTape` (hit when a `RoundTrip` races with `Close()`) now calls `onError` so data loss is never invisible, matching the existing behaviour of the buffer-full path.

- **Checked `runProxy`** — already safe: `defer tapeProxy.Close()` was already in place.

- **Checked `runServe`, `runExport`, `runImport`, `runMigrateFixtures`** — no async recorder, no changes needed.

## Test plan

- `TestRecordPersistsBufferedTapesOnInterrupt` (new) — starts the record command against a fake upstream, makes a probe request, sends SIGINT, waits for `run()` to return, then asserts the fixture is present *immediately* with no additional polling. This is the exact scenario that was broken before the fix.
- Updated comment in `runLiveCommandAndCapture` to reflect that the fixture-poll is now a no-op safety net (fixture is on disk when `done` unblocks).

Verified locally:
```
go build ./...
go vet ./...
go test ./... -race -count=1
go test -run TestRecordPersistsBufferedTapesOnInterrupt ./cmd/httptape/ -race -count=5
```
All green (5/5 for the new test).

### DIFF
--- a/cmd/httptape/main.go
+++ b/cmd/httptape/main.go
@@ -519,6 +519,18 @@ func runRecord(args []string) error {
 	}
 
 	recorder := httptape.NewRecorder(store, recorderOpts...)
+	// Drain async recordings synchronously before runRecord returns so that
+	// buffered tapes are never dropped on SIGINT. Without this defer,
+	// ListenAndServe returns (triggering run() exit) before the signal
+	// goroutine reaches recorder.Close(), losing all queued tapes. Close is
+	// idempotent via sync.Once, so the goroutine's Close (if reached) is safe.
+	defer func() {
+		if err := recorder.Close(); err != nil {
+			logger.Printf("recorder close error: %v", err)
+		} else {
+			logger.Println("recorder flushed")
+		}
+	}()
 
 	proxy := &httputil.ReverseProxy{
 		Rewrite: func(r *httputil.ProxyRequest) {
@@ -568,11 +580,7 @@ func runRecord(args []string) error {
 			logger.Printf("graceful shutdown failed: %v, forcing close", err)
 			httpServer.Close()
 		}
-		if err := recorder.Close(); err != nil {
-			logger.Printf("recorder close error: %v", err)
-		} else {
-			logger.Println("recorder flushed")
-		}
+		// recorder.Close() is handled by the defer above; no call needed here.
 	}()
 
 	scheme := "http"
--- a/recorder.go
+++ b/recorder.go
@@ -448,7 +448,10 @@ func (r *Recorder) persistTape(ctx context.Context, tape Tape) {
 		r.sendMu.Lock()
 		if r.closed.Load() {
 			r.sendMu.Unlock()
-			// recorder closed -- drop tape silently
+			// recorder already closed; notify via onError so data loss is visible
+			if r.onError != nil {
+				r.onError(fmt.Errorf("httptape: recorder closed, tape dropped"))
+			}
 		} else {
 			select {
 			case r.tapeCh <- tape:
--- a/cmd/httptape/main_test.go
+++ b/cmd/httptape/main_test.go
@@ -1603,14 +1603,12 @@ func runLiveCommandAndCapture(
 		t.Fatal("CLI did not exit after SIGINT")
 	}
 
-	// The recorder's Close() (which flushes async writes to disk) runs in the
-	// shutdown goroutine after httpServer.Shutdown returns. Because
-	// ListenAndServe returns as soon as Shutdown signals it, run() exits before
-	// the flush goroutine finishes. Poll until a fixture file appears so we can
-	// be confident the flush — and its final logger write ("recorder flushed")
-	// — has completed before we close the pipe and restore logger. Closing the
-	// pipe while the goroutine still holds a reference to the redirected logger
-	// would be a data race.
+	// After the fix for issue #304, recorder.Close() is called synchronously via
+	// defer in runRecord before run() returns. The fixture is therefore on disk
+	// when <-done unblocks. The short poll below is a safety net for proxy tests
+	// (where tapeProxy.Close is also deferred) and for any edge case where the
+	// OS write buffer is not yet visible. It exits in the first iteration when
+	// the fixture is already present.
 	flushDeadline := time.Now().Add(3 * time.Second)
 	for {
 		entries, _ := os.ReadDir(fixturesDir)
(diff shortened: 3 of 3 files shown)


cmd/httptape/main.go
```diff
@@ -568,11 +580,7 @@ func runRecord(args []string) error {
 			logger.Printf("graceful shutdown failed: %v, forcing close", err)
 			httpServer.Close()
 		}
-		if err := recorder.Close(); err != nil {
-			logger.Printf("recorder close error: %v", err)
-		} else {
-			logger.Println("recorder flushed")
-		}
+		// recorder.Close() is handled by the defer above; no call needed here.
```
tibtof, line 583: **Non-blocking correctness note — shutdown/Close ordering vs. in-flight requests.**

`recorder.Close()` now runs from the deferred func in `runRecord`, which fires as soon as `ListenAndServe` returns. But `http.Server.Shutdown` makes `ListenAndServe` return the moment it *closes the listener* — it then keeps blocking (in this fire-and-forget goroutine) until active connections drain. Nothing makes the main flow wait for `Shutdown` to finish before the deferred `Close` runs.

Consequence: a request already in-flight when SIGINT arrives can call `persistTape` (from `RoundTrip`) *after* `Close` has closed `tapeCh` → its tape hits the post-close drop path. This is serialized by `sendMu`/`closed` so there is no send-on-closed-channel panic (good), but the tape is lost.

Why it is not fatal here: (1) the exact scenario in #304 — tapes already **buffered** at SIGINT — is fully fixed, and the new test covers it deterministically; (2) any residual drop is now surfaced via `onError`. So this is strictly better than before (which dropped everything).

Why I am still flagging it: the *old* code had the correct ordering (Shutdown fully drains → then Close), it just never survived `os.Exit`. The clean fix that keeps that ordering AND survives exit is to have the main flow wait for the shutdown goroutine before returning — e.g. `shutdownDone := make(chan struct{})`, `close(shutdownDone)` at the end of the goroutine, and `<-shutdownDone` after `ListenAndServe` returns, keeping `Close` after that. Worth considering as a follow-up; not blocking the #304 fix.

recorder.go
```diff
@@ -448,7 +448,10 @@ func (r *Recorder) persistTape(ctx context.Context, tape Tape) {
 		r.sendMu.Lock()
 		if r.closed.Load() {
 			r.sendMu.Unlock()
-			// recorder closed -- drop tape silently
+			// recorder already closed; notify via onError so data loss is visible
+			if r.onError != nil {
+				r.onError(fmt.Errorf("httptape: recorder closed, tape dropped"))
```
tibtof, line 453: **onError on post-close drop — behavior/doc note (non-blocking).** Good change; it removes the silent data loss and mirrors the buffer-full path. Two small things:

1. **Semver-visible behavior change.** Library users who set `WithOnError` will now receive an extra `"recorder closed, tape dropped"` callback they never got before. This is additive/benign, but it is a visible behavior change on the error channel — worth a CHANGELOG line and it should land in a minor, not a patch.

2. **Doc drift.** `WithOnError`'s godoc says the callback is *"called from the background goroutine"*. This call (like the pre-existing buffer-full one) fires from the caller's `RoundTrip` goroutine. The "safe for concurrent use" note already covers correctness, but consider tightening the godoc to "may be called from `RoundTrip` and the background drain goroutine" so the concurrency contract is accurate.

Nit: you correctly release `sendMu` *before* invoking `onError` here, whereas the buffer-full path calls `onError` while still holding `sendMu` (line 461). Your ordering is the safer one — a follow-up could align the buffer-full path to unlock first, so a user callback that touches the recorder can't deadlock.

tibtof:
Reviewer verdict: **APPROVE** (posting as a comment — GitHub blocks a formal approval on ones own PR).

Reviewed the #304 fix end to end.

**Verified**
- **Drain fix correct.** `defer recorder.Close()` in `runRecord` runs on every exit path (signal, listener error, normal), LIFO after `defer stop()`. `Close` is idempotent via `sync.Once`, so the removed goroutine call was genuinely redundant. Buffered tapes are drained synchronously before `run()` returns.
- **No new deadlock.** `Close` waits on `<-r.done`; drain only does local `store.Save`. The signal goroutine no longer depends on `Close`. Slow-store hang is a pre-existing documented TODO, not introduced here.
- **onError post-close path** nil-guarded, invoked after `sendMu` release, mirrors the buffer-full path. Inline notes left re: semver/CHANGELOG visibility and a godoc-drift tightening.
- **runProxy/runServe/runExport/runImport unaffected** — proxy already deferred `tapeProxy.Close()`; others have no async recorder.
- **New regression test deterministic** — probe request fully returns (tape enqueued happens-before response) before SIGINT; fixture asserted immediately after `<-done`, no polling. `runLiveCommandAndCapture` only had comments changed; poll/settle logic intact, so no test weakened.

**Local runs**
- `go vet ./...` clean
- `go test ./... -race -count=1` green
- `go test ./cmd/... -race -count=5 -run Interrupt` green (5/5)

**One substantive non-blocking finding** (inline on `main.go:583`): `Shutdown` runs fire-and-forget and `ListenAndServe` returns as soon as the listener closes, so the deferred `Close` can run before `Shutdown` finishes draining. A request in-flight *at the instant of SIGINT* can still lose its tape (no panic — serialized by `sendMu`/`closed`; now surfaced via `onError`). Out of scope for #304 and strictly better than the prior all-or-nothing loss → not a merge blocker. Recommend a follow-up that waits on a `shutdownDone` channel before `Close` to restore the Shutdown→Close ordering.

No blocking issues. Approving.

Issue #310, opened 2026-06-11
config+docs: redact_query/fake_query rejected by config.schema.json, absent from docs; legacy actions accept stray params


tibtof:
## Summary
PR #295 added `RedactQueryParams`/`FakeQueryParams`/`DefaultSensitiveQueryParams` + config actions `redact_query`/`fake_query` (ADR-48) — the sanitize-on-write security boundary's newest surface — and it is documented **nowhere user-facing**. Repo-wide grep across docs/, README, llms*, doc.go, CHANGELOG returns zero hits. Two artifacts are factually wrong, one breaks CI:

### 1. `config.schema.json` REJECTS valid configs using the new actions (verified)
The rule `oneOf` (`config.schema.json:99-139`) has only `redact_headers`/`redact_body`/`fake` branches, each `additionalProperties: false` — `{"action":"redact_query"}` matches no branch and fails validation, while `LoadConfig` accepts it. `docs/config.md:271,277` advertises this schema for IDE autocompletion and CI validation, so CI pipelines validating against it hard-block the new security actions.
**Fix:** add two `oneOf` branches (`redact_query`: optional `params`; `fake_query`: required `seed` + non-empty `params`).

### 2. Legacy actions silently accept the new `params` field
`redact_query`/`fake_query` rules reject irrelevant fields, but `redact_headers`/`redact_body`/`fake` were not updated to reject `params` (`config.go:161-216`): `{"action":"redact_headers","params":["api_key"]}` validates cleanly and `params` is silently ignored — a user believing that rule covers query params will persist those secrets. Exactly the misconfiguration class the existing "does not use X" checks exist for.
**Fix:** add `params` rejection to the three legacy action cases in `Validate()`.

### 3. Docs gaps (factually wrong spots, not just omissions)
- `docs/config.md:248` — "known action (`redact_headers`, `redact_body`, `fake`)" — now false (five actions)
- `docs/api-reference.md:479-484` — Action constants block lists 3 of 5; `:242-249` sanitize-funcs table missing the two new funcs; `:287` has `DefaultSensitiveHeaders` but not `DefaultSensitiveQueryParams`
- `docs/sanitization.md` — per-function sections for every SanitizeFunc except the two new ones; no mention of always-on userinfo stripping, fail-closed malformed-query redaction, or fragment coverage (security-relevant, per ADR-48)
- `docs/config.md:98-137` — action reference missing `redact_query`/`fake_query` subsections
- llms.txt/llms-full.txt mirror all of the above

[...]

PR #343, opened 2026-08-05
chore(#311): v0.14.0 release prep — CHANGELOG + doc drift

### LINKED ISSUE #311: docs: post-#294/#295/#297 drift — single-flight key description, CHANGELOG, doc.go Sanitization section
## Summary
Smaller doc drift from the #294/#295/#297 wave, grouped for one PR:

### 1. Stale single-flight key description (contradicts ADR-47)
`docs/caching-transport.md:122`: "concurrent requests with the same match key **(method + path + body hash)** share a single upstream call" — describes exactly the unsafe pre-#294 keying. Actual key: method + path + raw query + body hash + canonical header hash minus denylist. Replace the parenthetical with the godoc one-liner and note the correctness-over-collapse trade-off. (The `WithCacheMatcher` default descriptions nearby are still correct — only this sentence drifted.)

### 2. CHANGELOG `[Unreleased]` missing #294/#295/#297
All three are user-facing (CLI default behavior change + new flag; new exported API + config actions; changed dedup semantics). Three bullets.

### 3. `doc.go` has no Sanitization section
Package overview covers Core types/Recording/Replay/Storage/Matching/Health — but not sanitization, the project's headline differentiator, and #295's new exports widened the gap. Add a `# Sanitization` section at principles altitude (Pipeline/SanitizeFunc, header/body/URL surfaces, deterministic faking, on-write boundary).

_Source: audit:2026-06 (2026-06-11 docs + architecture passes)._

tibtof:
Closes #311

## Summary

- **CHANGELOG**: converts `[Unreleased]` → `[0.14.0] - 2026-08-05` with a fresh empty `[Unreleased]` above. Fills all user-facing changes since v0.13.1 that were missing: #294 ADR-47 single-flight key fix (Fixed), #295 `RedactQueryParams`/`FakeQueryParams`/`DefaultSensitiveQueryParams` (Added), #297 CLI fail-closed + `--unsafe-raw` (Security + Added), #300/#335 SSE upstream-timeout fix (Fixed), #303/#340 `body_encoding: "base64"` marker (Added + Fixed), #304/#341 SIGINT drain + `WithOnError` racing-drop path (Fixed), #306/#336 matcher-only config fail-closed (Security + Fixed), #310/#337 config schema for `redact_query`/`fake_query` (Fixed). Consolidates the two duplicate "Breaking Changes" blocks from the prior draft.

- **doc.go**: adds a `# Sanitization` section covering `Pipeline`/`SanitizeFunc`, the four sanitization surfaces (headers, body JSON paths, URL query params + userinfo, SSE event data), HMAC-SHA256 deterministic faking, and the sanitize-on-write boundary. All cross-links point to real exported names.

- **docs/caching-transport.md line 122**: replaces the stale `(method + path + body hash)` parenthetical with the ADR-47 key (method + path + raw query string + body hash + canonical header hash minus denylist) and notes the correctness-over-collapse trade-off.

- **recorder.go `WithOnError` godoc**: clarifies that the callback may fire from the background drain goroutine _or_ from the `RoundTrip` goroutine when a tape is dropped on the racing-close path (added in #341). Code unchanged.

- **llms.txt + llms-full.txt**: fixes stale `NewMemoryStore()` → `NewMemoryStore(opts ...MemoryStoreOption)` in the Key APIs section.

## Test plan

- [x] `go vet ./...` — clean
- [x] `go test ./... -race -count=1` — green (both library and CLI packages)
- [x] `go build ./...` — clean
- [x] All godoc cross-links in doc.go verified against actual exported names in sanitizer.go
- [x] `git log v0.13.1..origin/main --oneline` cross-referenced against CHANGELOG entries — no user-facing PR omitted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+## [0.14.0] - 2026-08-05
+
 ### Added
 
 - **Inbound TLS listener**: CLI commands `serve`, `record`, and `proxy`
@@ -86,19 +88,82 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - **Config support for `path_pattern`**: declarative `"type": "path_pattern"`
   criterion with `"pattern"` field. (#196)
 
+- **`RedactQueryParams` / `FakeQueryParams`**: `SanitizeFunc` constructors
+  that redact or deterministically fake URL query parameter values. Userinfo
+  (`user:password@` in the authority) is always stripped, regardless of
+  which parameter names are configured. Both functions default to
+  `DefaultSensitiveQueryParams()` when called with no names. Fail-closed:
+  malformed percent-encoded query strings are replaced wholesale with
+  `[REDACTED]` rather than passed through as cleartext. (#295)
+
+- **`DefaultSensitiveQueryParams`**: returns a copy of the built-in list of
+  query parameter names commonly carrying sensitive data (`api_key`,
+  `access_token`, `token`, `secret`, `password`, `sig`, `signature`,
+  `X-Amz-Signature`, `X-Goog-Signature`). (#295)
+
+- **`--unsafe-raw` CLI flag**: `record` and `proxy` commands now apply safe
+  default sanitization (redact default sensitive headers + query params) unless
+  `--unsafe-raw` is passed. This flag disables all sanitization for callers who
+  explicitly opt out. The previous behavior (no sanitization by default) is
+  now opt-in. (#297)
+
+- **`body_encoding: "base64"` field**: non-UTF-8 text bodies that cannot be
+  stored as plain JSON strings are persisted as base64-encoded strings with an
+  explicit `"body_encoding": "base64"` marker. Fixtures written before this
+  change are unaffected (all-UTF-8 text bodies remain plain strings). (#303)
+
+### Security
+
+- **CLI fail-closed sanitization**: `httptape record` and `httptape proxy` now
+  default to redacting `DefaultSensitiveHeaders()` and `DefaultSensitiveQueryParams()`
+  from every recorded tape. Raw recording requires an explicit `--unsafe-raw` flag.
+  This eliminates the risk of committing API keys or session tokens to version
+  control when using the CLI without a config file. (#297)
+
+- **Matcher-only config fail-closed**: when a config file supplies a `matcher`
+  block but no `sanitizer` block, the CLI now rejects the invocation with an
+  error rather than silently recording without sanitization. A sanitizer is
+  required whenever a custom matcher is provided. (#306)
+
+### Fixed
+
+- **Single-flight key correctness**: the deduplication key for concurrent cache
+  misses now includes the raw query string and a canonical SHA-256 hash of all
+  request headers outside the hop-by-hop denylist. Previously, two requests
+  with the same method, path, and body but different query parameters or headers
+  could incorrectly share one upstream call and receive the same response.
+  (ADR-47, #294)
+
+- **SSE streams past upstream timeout**: SSE responses are now kept alive past
+  the `WithCacheUpstreamTimeout` deadline. The timeout previously applied to
+  the entire SSE stream duration; it now bounds only the header phase. (#300)
+
+- **Fail-closed on matcher-only configs**: the CLI now rejects a config file
+  that provides a `matcher` block without a `sanitizer` block rather than
+  recording without sanitization. (#306)
+
+- **Config schema for `redact_query` / `fake_query` actions**: the JSON config
+  schema now accepts `"redact_query"` and `"fake_query"` as sanitizer action
+  types. Legacy config actions that carry unrecognized fields now return an
+  error at load time rather than silently ignoring them. (#310)
+
+- **Non-UTF-8 text bodies persisted correctly**: text-typed response bodies
+  containing non-UTF-8 byte sequences (e.g. Latin-1 encoded text) now survive
+  the record → persist → replay round-trip byte-identically. Previously they
+  were silently corrupted by JSON encoding. (#303)
+
+- **SIGINT drains pending recordings**: `httptape record` and `httptape proxy`
+  now call `Recorder.Close()` before exiting on SIGINT, flushing any tapes
+  buffered in the async channel. Previously, tapes in flight at shutdown time
+  were silently dropped. (#304)
+
+- **`WithOnError` fires on racing-drop path**: `WithOnError` callbacks are now
+  also invoked when a tape is dropped because `RoundTrip` raced against
+  `Recorder.Close()`. Previously, only background drain-goroutine errors
+  triggered the callback; silent tape drops on the hot path were invisible to
+  callers. (#304)
+
 ### Breaking Changes
-
-- **`ResolveTemplateBody` and `ResolveTemplateHeaders` signatures changed**:
-  These now accept `*templateCtx` (unexported) instead of `*http.Request`.
-  External callers should use `ResolveTemplateBodySimple` instead. Pre-1.0,
-  acceptable. (#196)
-
-- **Unknown template namespaces**: expressions like `{{state.counter}}` that
-  were previously left as literal text are now replaced with empty string in
-  lenient mode (error in strict mode). All supported expressions are now
-  explicitly dispatched. (#196)
-
-### Breaking Changes (prior)
 
 - **`NewServer` signature change**: `NewServer(store Store, opts ...ServerOption)`
   now returns `(*Server, error)` instead of `*Server`. The constructor validates
@@ -116,10 +181,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   now returns `(SSETimingMode, error)` instead of `SSETimingMode`. Returns an
   error when factor is <= 0 instead of panicking. (#215, ADR-46)
 
+- **`ResolveTemplateBody` and `ResolveTemplateHeaders` signatures changed**:
+  These now accept `*templateCtx` (unexported) instead of `*http.Request`.
+  External callers should use `ResolveTemplateBodySimple` instead. Pre-1.0,
+  acceptable. (#196)
+
+- **Unknown template namespaces**: expressions like `{{state.counter}}` that
+  were previously left as literal text are now replaced with empty string in
+  lenient mode (error in strict mode). All supported expressions are now
+  explicitly dispatched. (#196)
+
 ### Migration
 
-All three changes are caught by the Go compiler -- no silent breakage. Update
-call sites as follows:
+All `NewServer` / `NewProxy` / `SSETimingAccelerated` changes are caught by
+the Go compiler — no silent breakage. Update call sites as follows:
 
 ```go
 // Before
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,30 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [0.14.0] - 2026-08-05
 
+### Module path change (breaking -- first release under the new org)
+
+v0.14.0 is the first tagged release published from the `httptape` GitHub
+org. v0.13.1 and earlier were tagged under `VibeWarden`. The Go module
+path, the `testcontainers` submodule path, and the container image
+namespace all changed:
+
+| | Before (≤ v0.13.1) | After (v0.14.0+) |
+|---|---|---|
+| Module path | `github.com/VibeWarden/httptape` | `github.com/httptape/httptape` |
+| `testcontainers` submodule path | `github.com/VibeWarden/httptape/testcontainers` | `github.com/httptape/httptape/testcontainers` |
+| Container image | `ghcr.io/vibewarden/httptape` | `ghcr.io/httptape/httptape` |
+
+**Migration:** update `go.mod` (and the `testcontainers/go.mod` submodule,
+if used) and every import site, then re-tidy:
+
+```bash
+grep -rl 'github.com/VibeWarden/httptape' --include='*.go' . go.mod | \
+  xargs sed -i '' 's#github.com/VibeWarden/httptape#github.com/httptape/httptape#g'
+go mod tidy
+```
+
+(#249, #250)
+
 ### Added
 
 - **Inbound TLS listener**: CLI commands `serve`, `record`, and `proxy`
@@ -91,10 +115,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - **`RedactQueryParams` / `FakeQueryParams`**: `SanitizeFunc` constructors
   that redact or deterministically fake URL query parameter values. Userinfo
   (`user:password@` in the authority) is always stripped, regardless of
-  which parameter names are configured. Both functions default to
-  `DefaultSensitiveQueryParams()` when called with no names. Fail-closed:
-  malformed percent-encoded query strings are replaced wholesale with
-  `[REDACTED]` rather than passed through as cleartext. (#295)
+  which parameter names are configured. `RedactQueryParams` defaults to
+  `DefaultSensitiveQueryParams()` when called with no names; `FakeQueryParams`
+  has no default -- called with no names, it fakes nothing (callers must pass
+  explicit parameter names). Fail-closed: malformed percent-encoded query
+  strings are replaced wholesale with `[REDACTED]` rather than passed through
+  as cleartext. (#295)
 
 - **`DefaultSensitiveQueryParams`**: returns a copy of the built-in list of
   query parameter names commonly carrying sensitive data (`api_key`,
@@ -111,6 +137,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   stored as plain JSON strings are persisted as base64-encoded strings with an
   explicit `"body_encoding": "base64"` marker. Fixtures written before this
   change are unaffected (all-UTF-8 text bodies remain plain strings). (#303)
+
+- **`redact_query` / `fake_query` config actions**: declarative JSON config
+  `rules` now support `"action": "redact_query"` and `"action": "fake_query"`,
+  mapping to `RedactQueryParams` / `FakeQueryParams`. The new `Rule.Params`
+  field lists the URL query parameter names to sanitize. (#310)
 
 ### Security
 
@@ -129,6 +160,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   is returned. (`serve` is unaffected — it relies on matcher-only configs
   legitimately.) (#306)
 
+- **Bundle import tar entry validation**: `ImportBundle` now rejects tar
+  entries with absolute paths, backslashes, or `..` path-traversal segments
+  before any entry is deserialized into a `Tape`. Defense in depth at the
+  bundle-parsing trust boundary for untrusted archives. (#218, #228)
+
 ### Fixed
 
 - **Single-flight key correctness**: the deduplication key for concurrent cache
@@ -138,14 +174,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   could incorrectly share one upstream call and receive the same response.
   (ADR-47, #294)
 
-- **SSE streams past upstream timeout**: SSE responses are now kept alive past
-  the `WithCacheUpstreamTimeout` deadline. The timeout previously applied to
-  the entire SSE stream duration; it now bounds only the header phase. (#300)
+- **SSE streams past upstream timeout**: `WithCacheUpstreamTimeout` previously
+  bounded the entire SSE stream duration, killing long-lived streams. It now
+  bounds only the header (time-to-first-byte) phase for SSE responses -- once
+  headers arrive, the timer is disarmed and the stream runs for its natural
+  lifetime. Buffered (non-SSE) responses are unaffected: the timeout still
+  covers the full upstream interaction, headers and body. (#300)
 
-- **Config schema for `redact_query` / `fake_query` actions**: the JSON config
-  schema now accepts `"redact_query"` and `"fake_query"` as sanitizer action
-  types. Legacy config actions that carry unrecognized fields now return an
-  error at load time rather than silently ignoring them. (#310)
+- **Config schema for `redact_query` / `fake_query` actions**: `config.schema.json`
+  now accepts `"redact_query"` and `"fake_query"` as sanitizer action types,
+  matching `Config.Validate`. See also the stray-`params` rejection for
+  legacy actions under Breaking Changes below. (#310)
 
 - **Non-UTF-8 text bodies persisted correctly**: text-typed response bodies
   containing non-UTF-8 byte sequences (e.g. Latin-1 encoded text) now survive
@@ -157,11 +196,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   buffered in the async channel. Previously, tapes in flight at shutdown time
   were silently dropped. (#304)
 
-- **`WithOnError` fires on racing-drop path**: `WithOnError` callbacks are now
-  also invoked when a tape is dropped because `RoundTrip` raced against
-  `Recorder.Close()`. Previously, only background drain-goroutine errors
-  triggered the callback; silent tape drops on the hot path were invisible to
-  callers. (#304)
+- **`WithOnError` fires on racing-drop path**: buffer-full drops and body
+  truncation notices already invoked `WithOnError` from the `RoundTrip`
+  goroutine. The one remaining silent path -- a tape dropped because
+  `RoundTrip` lost a race against `Recorder.Close()` -- now also invokes
+  `WithOnError`, so no tape drop is invisible to callers. (#304)
 
 ### Breaking Changes
 
@@ -180,6 +219,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - **`SSETimingAccelerated` signature change**: `SSETimingAccelerated(factor float64)`
   now returns `(SSETimingMode, error)` instead of `SSETimingMode`. Returns an
   error when factor is <= 0 instead of panicking. (#215, ADR-46)
+
+- **Legacy config actions now reject a stray `params` field**: `redact_headers`,
+  `redact_body`, and `fake` rules that include a `params` field now fail
+  `Config.Validate` (`"<action>" does not use "params"`) instead of silently
+  accepting and ignoring it. A config file that previously loaded
+  successfully with an errant `params` field on one of these actions will now
+  fail to load; remove the stray field or move it to a `redact_query` /
+  `fake_query` rule. (#310)
 
 - **`ResolveTemplateBody` and `ResolveTemplateHeaders` signatures changed**:
   These now accept `*templateCtx` (unexported) instead of `*http.Request`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,10 +120,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   This eliminates the risk of committing API keys or session tokens to version
   control when using the CLI without a config file. (#297)
 
-- **Matcher-only config fail-closed**: when a config file supplies a `matcher`
-  block but no `sanitizer` block, the CLI now rejects the invocation with an
-  error rather than silently recording without sanitization. A sanitizer is
-  required whenever a custom matcher is provided. (#306)
+- **Matcher-only / empty-rules config fail-closed**: when `--config` is
+  supplied but its `rules` array is empty or absent (e.g. a matcher-only
+  config), `record` and `proxy` now layer in the safe default sanitization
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -86,19 +88,82 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+  were silently dropped. (#304)
+
+- **`WithOnError` fires on racing-drop path**: `WithOnError` callbacks are now
+  also invoked when a tape is dropped because `RoundTrip` raced against
+  `Recorder.Close()`. Previously, only background drain-goroutine errors
+  triggered the callback; silent tape drops on the hot path were invisible to
+  callers. (#304)
 
-### Breaking Changes (prior)
+### Breaking Changes
```
tibtof, line 206: **Blocker — the biggest breaking change of v0.14.0 is missing.** Between `v0.13.1` and HEAD the Go module path was renamed (`f93b2e9`, #249/#250):

```
github.com/VibeWarden/httptape -> github.com/httptape/httptape
```

`git show v0.13.1:go.mod` still says `module github.com/VibeWarden/httptape`; `go.mod` on this branch says `github.com/httptape/httptape`. That means v0.14.0 is the first tag carrying the new path, and every embedder must edit their `go.mod` + every import site to upgrade. The same commit also moved the testcontainers submodule path (`github.com/httptape/httptape/testcontainers`) and the container image namespace (`ghcr.io/vibewarden/httptape` -> `ghcr.io/httptape/httptape`, now the `DefaultImage` in `testcontainers/httptape.go:16`), which breaks `docker pull` / compose files too.

Nothing in the `[0.14.0]` section mentions any of this. A user reading these notes will not learn why `go get github.com/VibeWarden/httptape@v0.14.0` fails.

Please add under Breaking Changes, and extend the Migration block:

```markdown
- **Module path renamed**: `github.com/VibeWarden/httptape` ->
  `github.com/httptape/httptape` (and `.../httptape/testcontainers` for the
  testcontainers submodule). The GitHub org migrated from `VibeWarden` to
  `httptape`. Update your `go.mod` require line and all import sites. (#249)

- **Container image namespace renamed**: `ghcr.io/vibewarden/httptape` ->
  `ghcr.io/httptape/httptape`. The Docker Hub namespace (`tibtof/httptape`)
  is unchanged. (#249)
```

CHANGELOG.md
```diff
@@ -86,19 +88,82 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - **Config support for `path_pattern`**: declarative `"type": "path_pattern"`
   criterion with `"pattern"` field. (#196)
 
-### Breaking Changes
+- **`RedactQueryParams` / `FakeQueryParams`**: `SanitizeFunc` constructors
+  that redact or deterministically fake URL query parameter values. Userinfo
+  (`user:password@` in the authority) is always stripped, regardless of
+  which parameter names are configured. Both functions default to
+  `DefaultSensitiveQueryParams()` when called with no names. Fail-closed:
```
tibtof: **Blocker — factually wrong, and wrong in the unsafe direction.** "Both functions default to `DefaultSensitiveQueryParams()` when called with no names" is true for `RedactQueryParams` only:

`sanitizer.go:234-236`
```go
func RedactQueryParams(names ...string) SanitizeFunc {
	if len(names) == 0 {
		names = DefaultSensitiveQueryParams()
	}
```

`sanitizer.go:297-303` — `FakeQueryParams` has no such fallback:
```go
func FakeQueryParams(seed string, names ...string) SanitizeFunc {
	// Build a set of query parameter names for O(1) lookup.
	sensitive := make(map[string]struct{}, len(names))
```

With no names the set is empty, so `FakeQueryParams(seed)` fakes nothing (userinfo is still stripped). `docs/sanitization.md:460` agrees with the code: for `fake_query` "the `params` field is required -- an empty list is rejected as a likely misconfiguration."

Release notes that tell a reader `FakeQueryParams(seed)` covers the default sensitive params will get secrets written to disk. Please split the sentence:

> `RedactQueryParams` defaults to `DefaultSensitiveQueryParams()` when called with no names; `FakeQueryParams` requires an explicit name list (an empty list is a no-op for parameters, though userinfo is still stripped).

CHANGELOG.md
```diff
@@ -86,19 +88,82 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+  were silently corrupted by JSON encoding. (#303)
+
+- **SIGINT drains pending recordings**: `httptape record` and `httptape proxy`
+  now call `Recorder.Close()` before exiting on SIGINT, flushing any tapes
+  buffered in the async channel. Previously, tapes in flight at shutdown time
+  were silently dropped. (#304)
+
+- **`WithOnError` fires on racing-drop path**: `WithOnError` callbacks are now
+  also invoked when a tape is dropped because `RoundTrip` raced against
+  `Recorder.Close()`. Previously, only background drain-goroutine errors
```
tibtof: **Inaccurate "previously" claim.** "Previously, only background drain-goroutine errors triggered the callback; silent tape drops on the hot path were invisible to callers" is not what the code did. `recorder.go:462-464` has fired `onError` from the `RoundTrip` goroutine on buffer-full drops since the original recorder commit (`110f824`), and the truncation notices at `recorder.go:328-337` / `390` / `422-423` do the same.

What #341 actually changed is narrower: the *recorder-closed* drop path (`recorder.go:453-455`) was the one silent drop. Suggest:

> Previously, a tape dropped on this path was silent — buffer-full drops and truncation already reported through `WithOnError`, but the closed-recorder race did not.

CHANGELOG.md
```diff
@@ -86,19 +88,82 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+- **Single-flight key correctness**: the deduplication key for concurrent cache
+  misses now includes the raw query string and a canonical SHA-256 hash of all
+  request headers outside the hop-by-hop denylist. Previously, two requests
+  with the same method, path, and body but different query parameters or headers
+  could incorrectly share one upstream call and receive the same response.
+  (ADR-47, #294)
+
+- **SSE streams past upstream timeout**: SSE responses are now kept alive past
+  the `WithCacheUpstreamTimeout` deadline. The timeout previously applied to
+  the entire SSE stream duration; it now bounds only the header phase. (#300)
```
tibtof: **Scope of the claim.** "it now bounds only the header phase" reads as unconditional, but that is SSE-only. `WithCacheUpstreamTimeout` godoc (`caching_transport.go:201-207`) is explicit:

> For buffered responses the timeout covers the entire upstream interaction: headers and body read. [...] For SSE responses the timeout bounds only the header (time-to-first-byte) phase.

Suggest: "...for SSE responses it now bounds only the header (time-to-first-byte) phase; buffered responses are still bounded end-to-end."

doc.go
```diff
@@ -82,6 +82,28 @@
 // With these options absent, [Proxy.HealthHandler] returns nil, no
 // goroutines are started, and proxy behavior is byte-for-byte unchanged.
 //
+// # Sanitization
+//
+// Sensitive data is redacted or faked before any tape touches disk —
+// there is no "raw recording" mode. Sanitization is applied via a
+// [Pipeline] of [SanitizeFunc] functions, each of which transforms a
+// [Tape] in place and returns it.
```
tibtof: **Blocker — contradicts the `SanitizeFunc` contract.** "transforms a [Tape] in place and returns it" is the opposite of what implementers must do. `sanitizer.go:313-317`:

```go
// SanitizeFunc is a function that transforms a Tape as part of a sanitization
// pipeline. Each function receives a Tape and returns a (possibly modified)
// copy. Implementations must not mutate the input Tape -- they must copy any
// fields they modify.
type SanitizeFunc func(Tape) Tape
```

Every built-in honours that (`RedactHeaders` clones the header maps, `FakeFields` copies the body slices, `RedactQueryParams` assigns a new URL string — each says so in its own godoc). This is the package-overview doc for a security boundary; a reader who writes a custom `SanitizeFunc` that mutates `t.Request.Headers` in place will silently corrupt the live request the caller still holds.

Suggest: "...applied via a [Pipeline] of [SanitizeFunc] functions, each of which receives a [Tape] and returns a modified copy — implementations never mutate the input."

doc.go
```diff
@@ -82,6 +82,28 @@
...
+//
+// Three surfaces are covered:
```
tibtof: **Count mismatch**: "Three surfaces are covered:" is followed by four bullets (headers, bodies, URL query params + userinfo, SSE event data). Change to "Four surfaces are covered:" — or drop the number entirely ("The following surfaces are covered:") so the list can grow without the prose drifting again.

doc.go
```diff
@@ -82,6 +82,28 @@
 // With these options absent, [Proxy.HealthHandler] returns nil, no
 // goroutines are started, and proxy behavior is byte-for-byte unchanged.
 //
+// # Sanitization
+//
+// Sensitive data is redacted or faked before any tape touches disk —
+// there is no "raw recording" mode. Sanitization is applied via a
```
tibtof: **Reintroduces framing that #338 removed in this same release.** "there is no \"raw recording\" mode" is no longer true on either side of the API:

- CLI: `--unsafe-raw` is exactly a raw recording mode (`cmd/httptape/main.go:441`, `514-515`), and `unsafeRawWarning` says "recording raw traffic with NO sanitization".
- Library: `NewRecorder(store)` without `WithSanitizer` records unsanitized — sanitization is opt-in for embedders.

#338 ("CLI sanitization is safe-by-default — fix opt-in framing everywhere") converged the docs on fail-closed wording, and this is the only place in the tree that still uses the absolutist phrasing (`grep -rn "raw recording" README.md docs/ doc.go` returns just this line). Compare `docs/sanitization.md:485` and `README.md:299`.

Suggest matching the shipped framing: "Sanitization happens on write, before a tape touches disk. The `record` and `proxy` CLI commands fail closed — a safe default pipeline is applied unless `--unsafe-raw` is passed; library embedders opt in with [WithSanitizer]."

CHANGELOG.md
```diff
@@ -86,19 +88,82 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+  (`user:password@` in the authority) is always stripped, regardless of
+  which parameter names are configured. Both functions default to
+  `DefaultSensitiveQueryParams()` when called with no names. Fail-closed:
+  malformed percent-encoded query strings are replaced wholesale with
+  `[REDACTED]` rather than passed through as cleartext. (#295)
+
+- **`DefaultSensitiveQueryParams`**: returns a copy of the built-in list of
+  query parameter names commonly carrying sensitive data (`api_key`,
+  `access_token`, `token`, `secret`, `password`, `sig`, `signature`,
+  `X-Amz-Signature`, `X-Goog-Signature`). (#295)
```
tibtof, line 129: **Missing from Added: the declarative config surface for #295.** The same PR shipped `ActionRedactQuery` / `ActionFakeQuery` constants and the `Rule.Params` field (`config.go:20-23`, `94`), i.e. `{"action": "redact_query"}` / `{"action": "fake_query", "params": [...]}` in config files. Right now that only surfaces obliquely in the Fixed bullet about the JSON *schema*, so a config-file user reading the release notes never learns the actions exist. Add a bullet:

```markdown
- **`redact_query` / `fake_query` config actions**: declarative equivalents of
  `RedactQueryParams` / `FakeQueryParams`, with a new `params` field on rules
  (`ActionRedactQuery`, `ActionFakeQuery`). (#295)
```

CHANGELOG.md
```diff
@@ -86,19 +88,82 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+  (ADR-47, #294)
+
+- **SSE streams past upstream timeout**: SSE responses are now kept alive past
+  the `WithCacheUpstreamTimeout` deadline. The timeout previously applied to
+  the entire SSE stream duration; it now bounds only the header phase. (#300)
+
+- **Config schema for `redact_query` / `fake_query` actions**: the JSON config
+  schema now accepts `"redact_query"` and `"fake_query"` as sanitizer action
+  types. Legacy config actions that carry unrecognized fields now return an
+  error at load time rather than silently ignoring them. (#310)
```
tibtof: **This one is breaking, not just Fixed.** "Legacy config actions that carry unrecognized fields now return an error at load time" means a config that loaded fine on v0.13.1 (e.g. a stray `params` on `redact_headers`) now fails `Validate()` at startup — `config.go` + `config_test.go` in `353a26c`. That is a silent-config-to-hard-failure transition on upgrade.

Keep the Fixed bullet, but cross-list it under Breaking Changes so upgraders know to re-validate their configs before rolling out.

CHANGELOG.md
```diff
@@ -86,19 +88,82 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+  control when using the CLI without a config file. (#297)
+
+- **Matcher-only / empty-rules config fail-closed**: when `--config` is
+  supplied but its `rules` array is empty or absent (e.g. a matcher-only
+  config), `record` and `proxy` now layer in the safe default sanitization
+  instead of proceeding with a no-op pipeline. A warning is printed to stderr
+  naming the redacted headers and query params and disclosing that
+  request/response bodies are not covered. The invocation proceeds; no error
+  is returned. (`serve` is unaffected — it relies on matcher-only configs
+  legitimately.) (#306)
```
tibtof, line 162: **Missing Security entry for #218/#228.** `b6048d7` added `validateEntryName` to `bundle.go`: `httptape import` now rejects tar entries with absolute paths, backslashes, or `..` traversal segments, returning `httptape: import: tar entry %q contains path traversal`. That is a user-visible behavior change (bundles that previously imported now error) on a defense-in-depth path, and it is absent from the whole `[0.14.0]` section. One bullet under Security:

```markdown
- **Bundle import rejects unsafe tar entry names**: `ImportBundle` / `httptape import`
  now reject entries with absolute paths, backslashes, or `..` traversal segments.
  Defense in depth — entries were already deserialized in-memory, never to the
  filesystem. (#218)
```

recorder.go
```diff
@@ -120,9 +120,11 @@ func WithBufferSize(size int) RecorderOption {
 	}
 }
 
-// WithOnError sets a callback invoked when an async store write fails.
-// The callback is called from the background goroutine, so it must be
-// safe for concurrent use. Defaults to a no-op (errors are discarded).
+// WithOnError sets a callback invoked when a store write fails or a tape is
+// dropped. The callback may be called from the background drain goroutine
+// (async write errors) or from the RoundTrip goroutine (e.g. when a tape is
+// dropped because the recorder was closed while RoundTrip was in flight). It
```
tibtof: Correct as far as it goes, but the godoc is still narrower than the code. `onError` also fires from the `RoundTrip` goroutine for buffer-full drops (`recorder.go:462-464`) and for body/SSE truncation notices (`328-337`, `390`, `422-423`) — the latter are not "a store write failed" nor "a tape was dropped" at all.

Since you are touching this comment anyway, consider making the first sentence cover the real contract:

```go
// WithOnError sets a callback invoked when the recorder cannot record
// faithfully: an async store write fails, a tape is dropped (buffer full,
// or the recorder was closed while RoundTrip was in flight), or a body is
// truncated at maxBodySize. It may be called from the background drain
// goroutine or from the RoundTrip goroutine, so it must be safe for
// concurrent use. Defaults to a no-op (errors are discarded).
```

Non-blocking, but the same "previously only the drain goroutine called it" misconception shows up in the CHANGELOG bullet for #304 — worth fixing both in one pass.

tibtof:

(Formal `--request-changes` is blocked — same GitHub account authored the PR. Verdict recorded here; 10 inline comments posted.)

Accuracy review of the highest-stakes docs in the repo. Five defects, three blocking.

### Verified clean

- `go vet ./...` clean; `go test ./... -race -count=1` green (both packages).
- Version mechanics: fresh empty `[Unreleased]`, `[0.14.0] - 2026-08-05`, format consistent with prior sections. This CHANGELOG has never carried link-reference definitions (`grep -n '^\[' CHANGELOG.md` is empty for every released section), and `.github/workflows/release.yml` uses `generate_release_notes: true` rather than parsing the file — nothing to update there.
- Breaking-changes list checked against current signatures, all four correct: `server.go:327` `NewServer(...) (*Server, error)`, `proxy.go:461` `NewProxy(...) (*Proxy, error)`, `sse.go:99` `SSETimingAccelerated(float64) (SSETimingMode, error)`, `templating.go:674`/`:732` taking `*templateCtx` with `ResolveTemplateBodySimple` at `:795`.
- `DefaultSensitiveQueryParams` enumeration matches `sanitizer.go:51-61` exactly — all nine names, same spelling and order.
- The #306 bullet matches `cmd/httptape/main.go:500-519` precisely: empty/absent `rules` layers in `defaultCLISanitizer()`, prints `emptyRulesConfigWarning` (bodies-not-covered disclosure included via `safeDefaultCoverage`), invocation proceeds with no error. The correction commit landed this right.
- `--unsafe-raw` / fail-closed wording matches post-#336 code and `README.md:299` / `docs/sanitization.md:485`.
- `docs/caching-transport.md:122` now matches the `WithCacheSingleFlight` godoc (`caching_transport.go:101-108`) near-verbatim, trade-off included. `llms-full.txt` carries only the condensed options table for that doc, not the prose sentence — no embedded copy left stale.
- `NewMemoryStore(opts ...MemoryStoreOption)` confirmed against `store_memory.go:23`.

### Blocking

1. **CHANGELOG omits the module path rename** (`f93b2e9`, #249/#250). `v0.13.1:go.mod` is `github.com/VibeWarden/httptape`; HEAD is `github.com/httptape/httptape`. v0.14.0 is the first tag carrying the new path, so every embedder must edit `go.mod` and all import sites — plus the testcontainers submodule path and the `ghcr.io/vibewarden/httptape` -> `ghcr.io/httptape/httptape` image namespace (`testcontainers/httptape.go:16`). The most impactful change in the release, entirely absent.

2. **The #295 bullet's default-params claim is wrong, in the unsafe direction.** `RedactQueryParams` falls back to `DefaultSensitiveQueryParams()` (`sanitizer.go:234-236`); `FakeQueryParams` does not (`sanitizer.go:297-303` builds the set from `names` only). `docs/sanitization.md:460` agrees with the code. As written, the notes tell a reader that `FakeQueryParams(seed)` covers the default sensitive params — it covers none of them.

3. **`doc.go:90` contradicts the `SanitizeFunc` contract.** "transforms a [Tape] in place and returns it" vs `sanitizer.go:313-317`: "Implementations must not mutate the input Tape -- they must copy any fields they modify." Every built-in documents its copying behavior. This is the package overview for the sanitization security boundary — it should not teach in-place mutation.

### Should fix in the same pass

4. `doc.go:92` says "Three surfaces" over four bullets.
5. `doc.go:88` "there is no \"raw recording\" mode" reintroduces the absolutist framing #338 removed earlier in this same release — `--unsafe-raw` is a raw recording mode, and library recorders without `WithSanitizer` record raw. It is now the only occurrence of that phrasing in the tree.
6. The #304 `WithOnError` bullet's "previously, only background drain-goroutine errors triggered the callback" is false — buffer-full drops (`recorder.go:462-464`) and truncation notices have fired from the `RoundTrip` goroutine since `110f824`. Only the closed-recorder race was silent. The same misconception is baked into the new `recorder.go` godoc.
7. The #300 bullet's "it now bounds only the header phase" is SSE-only; buffered responses are still bounded end-to-end (`caching_transport.go:201-207`).
8. Missing from Added: the `redact_query` / `fake_query` config actions and `Rule.Params` (`config.go:20-23`, `:94`) — currently only implied by a Fixed bullet about the JSON schema.
9. #310's stray-params rejection is breaking for existing config files (previously-loading configs now error at load time); cross-list under Breaking Changes.
10. #218/#228 bundle-import tar entry validation is a user-visible Security change and is absent.

### Drift item outside the diff

`docs/fixtures-authoring.md:99` claims `body_encoding: "base64"` was "introduced in v0.13+", but that line was added by #340 and ships in v0.14.0 — the claim is wrong the moment this tag is cut. This PR is the release-accuracy pass; fold it in.

tibtof:
All 10 inline comments + the verdict addressed in 92ba583:

**Blocking:**
1. Added a "Module path change" notice at the top of the [0.14.0] section — old/new module path, testcontainers submodule path, ghcr.io namespace, and sed migration (#249, #250).
2. Fixed #295 bullet: only `RedactQueryParams` defaults to `DefaultSensitiveQueryParams()`; `FakeQueryParams` fakes nothing with no names.
3. doc.go: removed "in place" mutation claim — now matches the SanitizeFunc contract (copy, don't mutate).

**Also fixed:**
4. doc.go "Three surfaces" → "Four surfaces".
5. doc.go: dropped "no raw recording mode" framing, replaced with safe-by-default wording consistent with #338.
6. Reworded the #304 WithOnError CHANGELOG bullet + recorder.go godoc — buffer-full/truncation already fired from RoundTrip; only the closed-recorder race was silent before #341.
7. #300 bullet: clarified header-phase-only bound is SSE-only; buffered responses stay bounded end-to-end.
8. Added CHANGELOG bullet for redact_query/fake_query config actions + Rule.Params.
9. Cross-listed #310's legacy-action stray-params rejection under Breaking Changes.
10. Added #218/#228 bundle-import tar-entry validation under Security.

Plus the out-of-diff drift item: docs/fixtures-authoring.md:99 now says v0.14.0 instead of "v0.13+".

`go vet ./...`, `go build ./...`, and `go test ./... -race -count=1` all green.

CHANGELOG.md
```diff
@@ -6,6 +6,32 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+| Module path | `github.com/VibeWarden/httptape` | `github.com/httptape/httptape` |
+| `testcontainers` submodule path | `github.com/VibeWarden/httptape/testcontainers` | `github.com/httptape/httptape/testcontainers` |
+| Container image | `ghcr.io/vibewarden/httptape` | `ghcr.io/httptape/httptape` |
+
+**Migration:** update `go.mod` (and the `testcontainers/go.mod` submodule,
+if used) and every import site, then re-tidy:
+
+```bash
+grep -rl 'github.com/VibeWarden/httptape' --include='*.go' . go.mod | \
+  xargs sed -i '' 's#github.com/VibeWarden/httptape#github.com/httptape/httptape#g'
```
tibtof: Nit (non-blocking) — `sed -i ''` is the BSD/macOS form. On GNU sed (Linux, most CI images) the empty string is consumed as the next argument and the command fails with `sed: can't read : No such file or directory`. A migration snippet in release notes gets copy-pasted into Linux CI a lot.

Portable on both:

```bash
grep -rl 'github.com/VibeWarden/httptape' --include='*.go' . go.mod | \
  xargs sed -i.bak 's#github.com/VibeWarden/httptape#github.com/httptape/httptape#g'
find . -name '*.bak' -delete
go mod tidy
```

Everything else in this section is verified correct: `v0.13.1:go.mod` is `github.com/VibeWarden/httptape`, HEAD is `github.com/httptape/httptape`, `testcontainers/go.mod` is `github.com/httptape/httptape/testcontainers`, and `testcontainers/httptape.go:16` `DefaultImage` is `ghcr.io/httptape/httptape:latest`. The table matches reality on all three rows.

CHANGELOG.md
```diff
@@ -86,19 +112,97 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+
+- **`body_encoding: "base64"` field**: non-UTF-8 text bodies that cannot be
+  stored as plain JSON strings are persisted as base64-encoded strings with an
+  explicit `"body_encoding": "base64"` marker. Fixtures written before this
+  change are unaffected (all-UTF-8 text bodies remain plain strings). (#303)
+
+- **`redact_query` / `fake_query` config actions**: declarative JSON config
+  `rules` now support `"action": "redact_query"` and `"action": "fake_query"`,
+  mapping to `RedactQueryParams` / `FakeQueryParams`. The new `Rule.Params`
+  field lists the URL query parameter names to sanitize. (#310)
```
tibtof: Nit (non-blocking) — attribution. The `redact_query` / `fake_query` actions, the `ActionRedactQuery` / `ActionFakeQuery` constants, and `Rule.Params` all shipped in #295 (`git log -S ActionRedactQuery -- config.go` -> `5242789`, "feat(sanitizer): redact/fake URL query params and userinfo (#295)"). #310/#337 only fixed `config.schema.json` and added the stray-`params` rejection. Suggest `(#295, #310)` or just `(#295)` so the Added bullet credits the PR that introduced the surface.

Content of the bullet is otherwise accurate against `config.go:20-23` and `:94`.

recorder.go
```diff
@@ -120,9 +120,12 @@ func WithBufferSize(size int) RecorderOption {
 	}
 }
 
-// WithOnError sets a callback invoked when an async store write fails.
-// The callback is called from the background goroutine, so it must be
-// safe for concurrent use. Defaults to a no-op (errors are discarded).
+// WithOnError sets a callback invoked when a store write fails or a tape is
+// dropped. The callback may be called from the background drain goroutine
+// (async write errors) or from the RoundTrip goroutine (buffer-full drops,
+// body-truncation notices, and -- since #341 -- the case where RoundTrip
```
tibtof: Nit (non-blocking) — "since #341" is a GitHub PR number in an exported symbol's godoc. On pkg.go.dev that renders as bare text with no link and no context for a reader who is not in this repo. The only other PR reference in non-test Go source is `caching_transport.go:47`, and that is an unexported struct-field comment.

Since this ships in v0.14.0, "since v0.14.0" carries the same information and is meaningful to a pkg.go.dev reader. The substance of the rewrite is correct — it now matches all three call-site classes (`recorder.go:328-337`/`390`/`422` truncation, `:462-464` buffer-full, `:453-455` post-close drop).

tibtof:
## Reviewer verdict: APPROVE

(Formal `--approve` is blocked — same GitHub account authored the PR. Verdict recorded here.)

Re-reviewed `92ba583` against all 11 findings from the first pass. Every one is fixed, and I re-verified each against the code rather than against the commit message.

### Blocking findings — all resolved

1. **Module path rename** — new `### Module path change (breaking -- first release under the new org)` section at the top of `[0.14.0]`. Verified all three table rows: `git show v0.13.1:go.mod` is `module github.com/VibeWarden/httptape`, HEAD is `github.com/httptape/httptape`, `testcontainers/go.mod` is `github.com/httptape/httptape/testcontainers`, and `testcontainers/httptape.go:16` `DefaultImage` is `ghcr.io/httptape/httptape:latest`. The "v0.13.1 and earlier were tagged under VibeWarden" claim holds — `git tag --sort=-creatordate` confirms v0.13.1 is the last pre-rename tag. Placement above `### Added` is right for something this consequential.

2. **`FakeQueryParams` default** — now reads "`RedactQueryParams` defaults to `DefaultSensitiveQueryParams()` when called with no names; `FakeQueryParams` has no default -- called with no names, it fakes nothing". Matches `sanitizer.go:234-236` vs `297-303` exactly, and the preceding sentence still correctly notes userinfo is stripped regardless. Agrees with `docs/sanitization.md:460`.

3. **`doc.go` `SanitizeFunc` contract** — "each receives a [Tape] and returns a (possibly modified) copy -- implementations must not mutate the input Tape" now mirrors `sanitizer.go:313-317` verbatim in substance.

### Should-fix findings — all resolved

4. "Four surfaces are covered:" over four bullets.
5. `doc.go:87-88` "there is no raw recording mode" replaced with sanitize-on-write framing, plus a new closing line naming the actual posture: opt-in via `[WithSanitizer]` for embedders (resolves to `recorder.go:78`), safe default in the CLI unless `--unsafe-raw`. Consistent with `README.md:299` and `docs/sanitization.md:485`.
6. `WithOnError` — both the CHANGELOG bullet and the godoc now state that buffer-full drops and truncation notices already fired from the `RoundTrip` goroutine, and scope #341 to the post-close race. Matches `recorder.go:328-337`/`390`/`422` (truncation), `:462-464` (buffer full), `:453-455` (post-close drop).
7. #300 bullet now scopes the header-phase bound to SSE and explicitly states buffered responses remain bounded end-to-end — matches `caching_transport.go:201-207`.
8. New Added bullet for the `redact_query` / `fake_query` config actions and `Rule.Params` (`config.go:20-23`, `:94`).
9. #310 stray-`params` rejection cross-listed under Breaking Changes, with the real error string `"<action>" does not use "params"` (`config.go:172`/`191`/`223`) and the upgrade consequence spelled out. The Fixed bullet now cross-references it instead of burying it.
10. #218/#228 bundle tar entry validation under Security — accurate to `bundle.go` `validateEntryName` (absolute paths, backslashes, `..` segments), and correctly notes it runs before deserialization.
11. `docs/fixtures-authoring.md:99` now says v0.14.0.

### Checks

- `go vet ./...` clean.
- `go test ./... -race -count=1` green — `github.com/httptape/httptape` and `.../cmd/httptape`.
- `[Unreleased]` still empty; `[0.14.0] - 2026-08-05`; section format consistent with prior releases; no link-reference block to maintain in this file.

### Three non-blocking nits (inline)

- `sed -i ''` in the migration snippet is BSD/macOS-only and fails on GNU sed; `sed -i.bak` + cleanup works on both.
- The new config-actions Added bullet is attributed to #310, but the actions, constants, and `Rule.Params` shipped in #295 (`5242789`); #310/#337 only fixed the schema and added the stray-`params` rejection.
- `recorder.go:126` cites "#341" in exported godoc, which renders as context-free text on pkg.go.dev; "since v0.14.0" says the same thing to that reader.

None of these change behavior or mislead about it. Fix them if you are pushing again before the tag; otherwise ready for human review and release.

